PR #7, opened 2016-09-13
Remove impossible case

aaronjensen:
According to dialyzer, this case is impossible. I would guess it has to do with what `lookup/1` is capable of returning. It's not capable of returning anything that would trigger the case of `extract_error_metadata/3` that would return nil.

```
endpoint.ex:53: The pattern {_, 'nil'} can never match the type {_,{atom(),binary(),_,_}}
```

It could be a dialyzer bug, there are a few nasty ones in 19.


arjan:
By just looking at the code, I see a case where `nil` is returned from `extract_error_metadata/3`, namely when the `plug_status < 500`. In that case we don't report an error because non-500 HTTP errors are not logged to AppSignal.

aaronjensen:
Fixed properly in #8 I believe.

PR #10, opened 2016-09-15
Fix remaining dialyzer error when used

aaronjensen:
This is a weird one, but basically because `extract_error_metadata` did not actually match all of the things that could be [sent to `reraise`](https://github.com/elixir-lang/elixir/blob/76b355a9d90989987174a87957622ecc372add00/lib/elixir/lib/kernel.ex#L1463-L1473) it gave these dialyzer errors:

```
lib/the_link/endpoint.ex:69: Guard test is_atom(_@4::map()) can never succeed
lib/the_link/endpoint.ex:69: Guard test is_binary(_@3::map()) can never succeed
```

I just added cases that behave like `reraise` does any way.

### DIFF
--- a/lib/appsignal/phoenix.ex
+++ b/lib/appsignal/phoenix.ex
@@ -57,6 +57,13 @@ defmodule Appsignal.Phoenix do
     # Submit error
     {r.__struct__, Exception.message(r), stack, Map.get(r, :conn, conn)}
   end
+  def extract_error_metadata(r, conn, stack) when is_binary(r) do
+    extract_error_metadata(RuntimeError.exception(r), conn, stack)
+  end
+  def extract_error_metadata(r, conn, stack) when is_atom(r) do
+    extract_error_metadata(r.exception([]), conn, stack)
+  end
+
 
   @doc false
   def submit_http_error(reason, message, stack, transaction, conn) do


arjan:
Thanks for this. Can you tell me how to get these errors from dializer, so I can add it to CI? I tried to run it  (`mix dialyzer`) but it only gave me warnings, no errors (e.g. a normal exit status 0)

lib/appsignal/phoenix.ex
```diff
@@ -57,6 +57,13 @@ defmodule Appsignal.Phoenix do
     # Submit error
     {r.__struct__, Exception.message(r), stack, Map.get(r, :conn, conn)}
   end
+  def extract_error_metadata(r, conn, stack) when is_binary(r) do
+    extract_error_metadata(RuntimeError.exception(r), conn, stack)
+  end
+  def extract_error_metadata(r, conn, stack) when is_atom(r) do
+    extract_error_metadata(r.exception([]), conn, stack)
```
arjan: I don't get `r.exception([])` here.. `r` is an atom at this point; calling a function on it will most likely fail, or not? Depending on whether `r` is a module..

aaronjensen:
You can get the errors by running dialyzer on a Phoenix project that uses appsignal. To get them in CI you would probably need to have a Phoenix project in an examples folder that pointed to appsignal via a path and run dialyzer/dialyxir on it. 

I copied the r.exception from reraise. I can't say I know when it happens, but I believe the assumption is that when it is an atom, it is a module that has an exception function.

aaronjensen:
Thanks for the merge and release!

arjan:
You're welcome :D

PR #22, opened 2016-10-11
Fix host metrics config key in GettingStarted

tgautier:
Very small typo in the GettingStarted guide but it seems, from [config.exs](https://github.com/appsignal/appsignal-elixir/blob/master/lib/appsignal/config.ex), that the key is `enable_host_metrics` and not `enable_host_metric`


arjan:
Thanks!

PR #26, opened 2016-10-19
remember original stacktrace in phoenix endpoint

manukall:
before, if an exception was raised during processing the transaction, the stacktrace for that exception was sent.

### DIFF
--- a/lib/appsignal/phoenix.ex
+++ b/lib/appsignal/phoenix.ex
@@ -31,7 +31,7 @@ defmodule Appsignal.Phoenix do
           e ->
             stacktrace = System.stacktrace
             import Appsignal.Phoenix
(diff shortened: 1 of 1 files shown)


lib/appsignal/phoenix.ex
```diff
@@ -29,14 +29,15 @@ defmodule Appsignal.Phoenix do
           super(conn, opts)
         rescue
           e ->
+            stacktrace = System.stacktrace
             import Appsignal.Phoenix
             case {Appsignal.TransactionRegistry.lookup(self), extract_error_metadata(e, conn, System.stacktrace)} do
```
arjan: Makes sense to use the `stacktrace` var here as well to prevent calling the `System.stacktrace` function twice.

manukall: absolutely. didn't see that before.

arjan:
Thanks!

PR #40, opened 2016-12-07
Send metadata in Appsignal.ErrorHandler.submit_transaction/6

### LINKED ISSUE #39: metadata sent via send_error does not show up in appsignal
Using `Appsignal.send_error/4`, I'm not seeing any of the metadata I send in the 4th argument in my error on appsignal. 

I'm using https://github.com/joakimk/toniq/issues/29#issuecomment-265251468

jeffkreeftmeijer:
Fixes #39.

### DIFF
--- a/lib/appsignal/error_handler.ex
+++ b/lib/appsignal/error_handler.ex
@@ -54,6 +54,7 @@ defmodule Appsignal.ErrorHandler do
     if metadata[:conn] != nil do
       Transaction.set_request_metadata(transaction, metadata[:conn])
     end
+    Transaction.set_meta_data(metadata)
     Transaction.finish(transaction)
     Transaction.complete(transaction)
     Logger.debug("Submitting #{inspect transaction}: #{message}")
--- a/lib/appsignal/error_handler.ex
+++ b/lib/appsignal/error_handler.ex
@@ -32,12 +32,12 @@ defmodule Appsignal.ErrorHandler do
   def handle_event(event, state) do
     state =
      case match_event(event) do
-        {origin, reason, message, stack, metadata, conn} ->
+        {origin, reason, message, stack, conn} ->
           case supervisor_event?(reason) do
             false ->
               transaction = Transaction.lookup_or_create_transaction(origin)
               if transaction != nil do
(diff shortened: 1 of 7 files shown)


lib/appsignal/transaction.ex
```diff
@@ -278,10 +278,11 @@ defmodule Appsignal.Transaction do
   """
   @spec set_meta_data(Transaction.t | nil, String.t, String.t) :: Transaction.t
   def set_meta_data(nil, _key, _value), do: nil
-  def set_meta_data(%Transaction{} = transaction, key, value) do
+  def set_meta_data(%Transaction{} = transaction, key, value) when is_binary(key) and is_binary(value) do
     :ok = Nif.set_meta_data(transaction.resource, key, value)
     transaction
   end
+  def set_meta_data(%Transaction{} = transaction, _key, _value), do: transaction
```
arjan: I'm wondering if this is the right approach, to silently ignore all metadata when key or value is not a string.
E.g. someone might assume the key is an atom; which is supported in the case of passing in a top-level map.

lib/appsignal/error_handler.ex
```diff
@@ -32,12 +32,12 @@ defmodule Appsignal.ErrorHandler do
   def handle_event(event, state) do
     state =
      case match_event(event) do
-        {origin, reason, message, stack, metadata} ->
+        {origin, reason, message, stack, metadata, conn} ->
```
arjan: Great, only I think that match_event always returns `%{}` for the metadata. So it could be left out entirely.

jeffkreeftmeijer: You're right. I removed the metadata from `match_event/1` in 9579739ae382eb586f184da7f5a81395292d3fb2. As you can see, now it always calls `submit_transaction/6` with an empty map. 

If we'd change the attribute order in `submit_transaction/6` to `transaction, reason, message, stack, conn // nil, metadata // %{}`, we would drop the empty map in the call from `handle_event/2`. However, then it would mismatch the attribute order in `Appsignal.send_error/5`, where it's the other way around. 

We could change that too, but I'm not sure we want to change the API there to prevent having to use an empty map in `handle_event/2`. What do you think?

PR #56, opened 2016-12-16
Use APPSIGNAL_APP_ENV instead of APPSIGNAL_ENVIRONMENT

jeffkreeftmeijer:
Use `APPSIGNAL_APP_ENV`, like in the [Ruby integration](http://docs.appsignal.com/ruby/configuration/options.html#code-appsignal_app_env-code), and set `APPSIGNAL_ENVIRONMENT` from `write_to_environment/1` using the `:env` config, for the agent. Based on #55.


arjan:
I see that you still keep the `APPSIGNAL_ENVIRONMENT` variable as well. Is this really needed? Otherwise, good to go.

arjan:
Oh, I get it now, the old one is not being read from.

PR #57, opened 2016-12-28
Added suport to refs and pids inside payloads

costaraphael:


jeffkreeftmeijer:
Awesome. Thanks @costaraphael! :+1:

PR #62, opened 2017-01-10
Add a transcation decorator

### LINKED ISSUE #52: Transaction decorator
To make it easier to instrument events outside of a transaction, we'll add a transaction decorator that starts a transaction itself. Here's a [prototype](https://github.com/jeffkreeftmeijer/appsignal-elixir-example/commit/5a4b4b348e9d166b65eb8b7721eea0f48169f73e).

jeffkreeftmeijer:
Closes #52.

### DIFF
--- a/lib/appsignal/instrumentation/decorators.ex
+++ b/lib/appsignal/instrumentation/decorators.ex
@@ -25,9 +25,32 @@ defmodule Appsignal.Instrumentation.Decorators do
 
 
   use Decorator.Define,
+    transaction: 0,
+    transaction: 1,
     transaction_event: 0,
     transaction_event: 1,
     channel_action: 0
+
+  @doc false
+  def transaction(body, context) do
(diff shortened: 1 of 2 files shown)


arjan:
Looks good, but I'm missing the documentation; can you add some to the `@moduledoc`?

arjan:
Looks good

I'm wondering; what's missing is the case when Appsignal needs to collect metrics (when `:sample` is returned from `Transaction.finish()`). Nothing is done right now. This is also hard because ideally we need the `%Plug.Conn{}` at that point to collect any sensible metadata:

```
    if Transaction.finish(transaction) == :sample do
      Transaction.set_request_metadata(transaction, conn)
    end
```

Maybe we should leverage ETS so that the inner function can attach any metadata (like the conn) to the transaction. But lets keep that for a later ticket.

Issue #53, opened 2016-12-15
Error running ecto data update migrations with Appsignal installed


navinpeiris:
I'm trying to run a migration which simply updates the values of a particular table column using Repo.update_all, but when Appsignal is active, it throws the following error:

```elixir
** (ArgumentError) argument error
    (stdlib) :ets.lookup(:"$appsignal_transaction_registry", #PID<0.73.0>)
    lib/appsignal/transaction/registry.ex:46: Appsignal.TransactionRegistry.lookup/2
    lib/appsignal/ecto.ex:23: Appsignal.Ecto.log/1
    (db_connection) lib/db_connection.ex:1177: DBConnection.log/6
    (ecto) lib/ecto/adapters/postgres/connection.ex:78: Ecto.Adapters.Postgres.Connection.prepare_execute/5
    (ecto) lib/ecto/adapters/sql.ex:235: Ecto.Adapters.SQL.sql_call/6
    (ecto) lib/ecto/adapters/sql.ex:424: Ecto.Adapters.SQL.execute_and_cache/7
    (stdlib) timer.erl:197: :timer.tc/3
```

I've got the `Appsignal.Ecto` logger configured via:

```elixir
config :hoffer, MyApp.Repo,
  loggers: [Appsignal.Ecto]
```

This is on:

```
elixir: 1.3.4
ecto: 2.0.6
appsignal: 0.9.2
```
The workaround for me with this was to remove the `Appsignal.Ecto` logger config so that the migration could run, and then re-enable it again.

Do let me know what other information might be helpful, I'll get them across as soon as possible.

jeffkreeftmeijer:
Hey @navinpeiris, thanks for reporting this (and sorry for the inconvenience)! I'll create an example app to reproduce this, and I'll get back to you as soon as I know more.

arjan:
@jeffkreeftmeijer  it seems the migrations are run when the `appsignal` application is not started. The fix would be to check whether the application is running inside the ecto logger. or maybe even in the transaction registry.

PR #70, opened 2017-01-13
Registry.register/1 returns nil if Appsignal is not started

jeffkreeftmeijer:


arjan:
Can you remind me again why there should be the case that Appsignal is not started?
As soon as you list `:appsignal` in your application's list, it will start. There should be no reason to stop it.

arjan:
Oh wait, #53.

jeffkreeftmeijer:
@arjan correct, #53 was the reason we're checking if AppSignal is started in `lookup/2`. It crashed on `register/1` yesterday when @tombruijn was working on implementing the mix task. In that case, you should start Appsignal manually.

PR #73, opened 2017-01-13
Add fix for grabbing filter_parameters from Phoenix

tylerpearson:
Enjoying AppSignal with Elixir so far, but I noticed an issue where our users' passwords are showing up in the parameters section on the website, which is no bueno. It looks like this is due to how the filter parameters are being grabbed from Phoenix. The [default parameters for Phoenix](https://github.com/phoenixframework/phoenix/blob/d88a34ec5d316d6f9379f63d41d149df259c9813/mix.exs#L42) live in `Application.get_env(:phoenix, :filter_parameters)` instead of `:config`, so this isn't properly grabbing the defaults and secrets are being sent.

Also, I'll submit a PR for a fix with the docs. For setting it with AppSignal's filtering, it *does* check `:config`, but the docs don't mention that.

### DIFF
--- a/lib/appsignal/utils/params_filter.ex
+++ b/lib/appsignal/utils/params_filter.ex
@@ -6,7 +6,7 @@ defmodule Appsignal.Utils.ParamsFilter do
 
   def get_filter_parameters do
     Application.get_env(:appsignal, :config)[:filter_parameters]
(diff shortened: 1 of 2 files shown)


arjan:
Looks good, thanks for this fix!

PR #75, opened 2017-01-16
Send all request headers

arjan:
They are visible under the "environment" section in the Appsignal UI.


arjan:
Looks like this:

image

PR #79, opened 2017-01-17
Add the module name to the transaction action while using decorators

costaraphael:
Sometimes, specially using background jobs libraries, the function names are always the same (like `perform`), while the module name differentiate the actions (`EmailDeliverWorker`, `CacheUpdateWorker`, ...).

This fix adds the module name to the action name, so it resembles more the action names of controllers.


arjan:
Good addition, thanks :)

Issue #80, opened 2017-01-17
Application is not compiling after update to 0.11.0


costaraphael:
When I try to compile my application for the first time, the following error happens:
![image](https://cloud.githubusercontent.com/assets/7742817/22039300/1f95acae-dcd5-11e6-9352-b6e0dfde979e.png)

Looking at the source, I found that there is a new check at the top of `lib/appsignal/phoenix.ex`, that checks if Phoenix is present with `Code.ensure_loaded?/1`.

The problem is, when the appsignal application is compiled, phoenix is not compiled yet, which causes the check to return false.

A workaround to this is to compile phoenix first and then compile the rest of the dependencies: `mix do deps.compile plug, deps.compile phoenix, compile`.

I guess that there are better ways to check if phoenix is present (maybe using some mix functions?).

arjan:
Hmm this is a clear oversight on our part. I thought i saw this method of compile-time checking in a different project as well, that's why we did it this way. I'll have a look to see if it can be done differently.

arjan:
In your app, do you list the `:phoenix` dependency before `:appsignal`?

I got the idea of doing it like this from Ecto [see here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/poison.ex#L1).

costaraphael:
But Ecto uses poison as a optional dependency, [look](https://github.com/elixir-ecto/ecto/blob/master/mix.exs#L52).

Reordering my deps is kind of confusing, since I have an umbrella application where one app called `API` depends on phoenix and another app called `Core`. `Core` is the one that depends on appsignal.

arjan:
Oh wait, maybe it has to do with the `only: :test_phoenix` clause in appsignal's phoenix dep.

arjan:
Yes, that's it. Will release 0.11.1 tomorrow with a fix, thanks for the headsup.

pdilyard:
Having the same issue, looking forward to the release

jeroenvisser101:
Same issue here as well, is there anything that I can change to make it compile? Can I force AppSignal to build with Phoenix?

PR #82, opened 2017-01-18
Prevent module not loaded errors in phoenix

jeroenvisser101:
Prevents "module Appsignal.Phoenix is not loaded and could not be found" errors when Appsignal is compiled before Phoenix and breaks the `Code.ensure_loaded?(Phoenix)` check.

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -18,6 +18,8 @@
       [applications: [:appsignal]]
     end
     ```
+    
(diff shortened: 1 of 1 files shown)


arjan:
Thanks. We've just fixed this in a different way, see #80, just pushed to master. Can you doublecheck our solution?

jeroenvisser101:
Ah nice, checking now!

PR #83, opened 2017-01-18
Readded support for pids and refs inside payloads

costaraphael:
This support used to exist, but was dropped when `ParamsEncoder` was replaced by `DataEncoder`.

I added a clause to specifically handle atoms (to maintain the `:data` -> `"data"` behavior), and switched the `to_string` to a `inspect` call in the catch-all clause.

### DIFF
--- a/lib/appsignal/utils/data_encoder.ex
+++ b/lib/appsignal/utils/data_encoder.ex
@@ -49,8 +49,11 @@ defmodule Appsignal.Utils.DataEncoder do
   def encode(resource, {key, nil}) do
     Nif.data_set_nil(resource, key)
   end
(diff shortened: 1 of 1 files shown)


jeffkreeftmeijer:
Hey @costaraphael, sorry for missing this. I've added a test to DataEncoder to make sure we don't break this again. Thanks for sending yet another pull request!

jeffkreeftmeijer:
Also, I just released 0.11.2 with this fix. Thanks again @costaraphael!

costaraphael:
You're welcome ^^

Thanks for the fast release (I was just pushing a commit into our repo to point the appsignal dep to the `develop` branch :joy:)

Issue #90, opened 2017-01-24
Don't crash when the NIF fails to load


jeffkreeftmeijer:

jeffkreeftmeijer:
Fixed in #93.

Issue #89, opened 2017-01-24
Don't crash the compilation on unsupported OS'es


arjan:
Currently the compilation fails when trying to compile Appsignal e.g. on windows (see #36). It would be nicer if the NIF compilation would just be skipped and a big red warning printed that Appsignal is not supported on the platform.

cc @nulian

jeffkreeftmeijer:
Fixed in #93.

arjan:
@nulian would you be so kind to test this on your windows machine? :)

PR #93, opened 2017-01-25
Don't crash when target platform is not supported

arjan:
This addresses both #89 and #90.

On compile time, a warning is printed when the CPU platform is not supported.

On runtime, while loading the NIF, an error is printed. All NIF
stub functions now return stub values and do not call exit() so that
the integration keeps working (without doing anything) when running on
a non-supported OS.

@jeffkreeftmeijer can you test this somehow?

### DIFF
--- a/lib/appsignal/nif.ex
+++ b/lib/appsignal/nif.ex
@@ -29,9 +29,17 @@ defmodule Appsignal.Nif do
 
   @on_load :init
 
+  require Logger
+
   def init do
     path = :filename.join(:code.priv_dir(:appsignal), 'appsignal_extension')
-    :ok = :erlang.load_nif(path, 1)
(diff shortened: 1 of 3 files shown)


arjan:
You can partially test this by returning `:unsupported` in `mix_helpers.exs` for all archs.
However, the JSON encoding tests will fail. I'm wondering whether we want to expose this `data_to_json` function in the Appsignal API at all..? People might misuse it as a generic JSON encoding mechanism.

jeffkreeftmeijer:
Tested this, as you suggested, by returning `:unsupported` from `Mix.Appsignal.Helpers.map_arch/1`. 

The `data_to_json()` function is specifically added to the agent to help with testing the C extension and Nif, and won’t be used elsewhere. You’re right, we shouldn’t export it in the Nif, at least.

PR #109, opened 2017-02-01
allow me to use poison 3.1.0

izelnakri:


jeffkreeftmeijer:
Hey @izelnakri! Thanks for bringing this to our attention and thanks for your pull request. Instead of loosening the dependency requirements, we found a way to get rid of our Poison dependency completely in #124. This should solve your problem, and saves us a dependency. Thanks again! :)

PR #110, opened 2017-02-02
C extension review

thijsc:
Rename `StringValueCStr` to `make_appsignal_string` and check that pointers returned from the Rust side are not null.


thijsc:
Closing this one, we've made a change on the Rust side to make sure a pointer is always returned.

PR #124, opened 2017-02-10
Use config/agent.exs instead of agent.json

### LINKED ISSUE #115: Remove Poison dependency
As discussed in https://github.com/appsignal/appsignal-elixir/pull/112#issuecomment-277252491, we're removing the agent.json file and will be replacing it with a config file we can read without having to depend on a JSON library, to fix #109.

jeffkreeftmeijer:
Closes #109 and closes #115.


izelnakri:
awesome, I liked it!

PR #187, opened 2017-03-27
Catch errors caught by Plug.ErrorHandler

jeffkreeftmeijer:
Instead of rescuing from exceptions and re-raising them using a rescue
block, this switches it over to a catch block, which uses
`Plug.ErrorHandler.__catch__/4` to re-raise the error. This allows us to
catch errors caused by crashing processes (like the timeout in the added
test).

This can be merged on its own, but not released without #190, as that could cause duplicate errors to be registered.

### DIFF
--- a/lib/appsignal/error_handler.ex
+++ b/lib/appsignal/error_handler.ex
@@ -201,6 +201,9 @@ defmodule Appsignal.ErrorHandler do
     msg = Exception.message(r)
     {"#{inspect r.__struct__}", prefixed(message, msg)}
   end
+  def extract_reason_and_message({kind, _} = reason, message) do
+    {inspect(kind), prefixed(message, inspect(reason))}
+  end
   def extract_reason_and_message(any, message) do
     # inspect any term; truncate it
     {"#{inspect any}", message}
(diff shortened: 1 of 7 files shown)


arjan:
> This drops support for unsupervised "bare" spawns
Isn't this something that people are going to miss?

jeffkreeftmeijer:
@arjan replied in https://github.com/appsignal/appsignal-elixir/pull/190#issuecomment-290068062, as that pull request includes the commit you're referring to.

#190 should be merged before this can be reviewed.

jeffkreeftmeijer:
From @tombruijn in https://github.com/appsignal/appsignal-elixir/pull/190#discussion_r108663173:

> the commit message of this commit should be "move" rather than "remove" I think?
Also the test you added earlier in the plug should also be added to the test for this module? (Or moved)

Updated to "move". Also, a test was updated to test this in https://github.com/appsignal/appsignal-elixir/pull/187/commits/231172d4a722e409111655bddfd71170513ee04e#diff-9e81d8c6df508df48af5797149975f56.

PR #198, opened 2017-04-10
Allow nil transactions

xymbol:
This change allows calls to `Appsignal.Instrumentation.Helpers.instrument/6` with `nil` as an argument. In a way, it tries to make `instrument/6` consistent with `instrument/3` when no transaction is found for the current process.

This helps avoiding clauses or conditionals in instrumented code. For example:

```elixir
defmodule HelloServer do
  use GenServer

  def handle_call({:hello, name, transaction}, _from, state) do
    instrument(transaction, "hello", "Handling hello request", fn ->
      {:reply, do_hello(name), state}
    end)
  end

  def handle_call({:hello, name, nil}, _from, state) do
    {:reply, do_hello(name), state}
  end

  defp do_hello(name) do
    "Hello, #{name}!"
  end
end
```

With this patch the second clause for `handle_call/3` can be dropped.

We ran into this issue in a production app where we need to report nested times for client and server processes. In order to do that, we pass in a transaction. But, in tests, instrumentation is not wanted and would rather avoid the duplication.

### DIFF
--- a/lib/appsignal/instrumentation/helpers.ex
+++ b/lib/appsignal/instrumentation/helpers.ex
@@ -5,7 +5,7 @@ defmodule Appsignal.Instrumentation.Helpers do
 
   alias Appsignal.{Transaction, TransactionRegistry}
 
-  @type instrument_arg :: Transaction.t | Plug.Conn.t | pid()
+  @type instrument_arg :: Transaction.t | Plug.Conn.t | pid() | nil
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Hey @xymbol! :wave: Thanks for looking into this, this sure is a lot cleaner. :)

PR #204, opened 2017-04-25
phoenix template: use path as title instead of name

kagux:
I've noticed that slow events report groups template render events from multiple controllers. That's not expected and I'd like to see event response time separately.


jeffkreeftmeijer:
Hey @kagux! :wave:

Great find, and thanks for taking the time to submit a pull request for this! I've added a quick and dirty regression test (so we don't break this) in 6bec2246081d37f45e14a96ff6ea958376adf68a.

Thanks again!

PR #205, opened 2017-04-25
Improve registry lookup performance

kagux:
Hi,

I'll start with a short background story so you can have a context of why this change is important.
We use quite a few partials in our project to render relatively big collections (100-500 items). Template rendering was taking up to 50% of time on such pages according to performance reports. 

We started adding more instrumentation to views in a mission to narrow down the bottleneck. This resulted in a noticeable increase of render time. We also saw performance reports pointing that the issue was in rather view helpers doing simple operations on small lists.

We used Benchfella to verify the issue with these view helpers and quickly discovered that it was caused by Appsignal decorated functions being invoked in a loop. Benchmarking further only Appsignal instrumentation we pinned the root cause to `Appsignal.started?` function that uses relatively slow `Application.get_applicatin/1`.

`Appsignal.stated?` was added as a fix to issue #53, where uninitialized ets table was accessed because appsignal was not started. This PR proposes to check if `Appsignal.TransactionRegistry` process is running instead. After all that's what we're really interested in, as ets table is coupled to this process. Checking for ets table presense could be another option, but checking for process presence seems to be more decoupled approach.

Anyway, here's the benchmark code

```
{:ok, _} = Application.ensure_all_started(:appsignal)

defmodule Appsignal.InstrumentationBench do
  use Benchfella
  import Appsignal.Instrumentation.Helpers, only: [instrument: 3]

  bench "instrumenting no-op function" do
    instrument("no-op", "no-op", fn() -> :ok end)
  end
end
```

Appsignal 1.2.0
```
benchmark name                iterations   average time
instrumenting no-op function     1000000   19.94 µs/op
```

PR branch
```
benchmark name                iterations   average time
instrumenting no-op function   100000000   0.31 µs/op
```

And a couple screenshots from performance reports after we switched to this fork with no other changes:

1. Template that uses ~ 30 partials (including nested) and view helpers.
image

2. View helper that used instrumented function in a loop
image

3. Controller using above template
image

### DIFF
--- a/lib/appsignal.ex
+++ b/lib/appsignal.ex
@@ -34,10 +34,6 @@ defmodule Appsignal do
     Supervisor.start_link(children, [strategy: :one_for_one, name: Appsignal.Supervisor])
   end
 
-  def started? do
-    Application.get_application(Appsignal) != nil
-  end
-
   def phoenix? do
     Code.ensure_loaded?(Phoenix)
   end
--- a/lib/appsignal/transaction/registry.ex
+++ b/lib/appsignal/transaction/registry.ex
@@ -34,13 +34,12 @@ defmodule Appsignal.TransactionRegistry do
   @spec register(Transaction.t) :: :ok
   def register(transaction) do
     pid = self()
-    case registry_alive?() do
-      true ->
-        true = :ets.insert(@table, {pid, transaction})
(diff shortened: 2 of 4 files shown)


arjan:
:+1: `application:get_application()` indeed uses a serialized request, great find.

PR #257, opened 2017-09-18
Allow custom category in Appsignal.Ecto

karolsluszniak:
This allows to change the default `ecto` category name for Ecto queries. This is super useful when multiple repos are used in a single project and during single transaction, as it allows to put those queries into distinct category buckets as well as easily differentiate them within a single transaction sample.

In order to use it, one has to change the default:

```elixir
config :my_app, MyApp.Repo,
  loggers: [Appsignal.Ecto, Ecto.LogEntry]
```
 
to something along these lines:

```elixir
config :my_app, MyApp.Repo,
  loggers: [{Appsignal.Ecto, :log, [:ecto_repo]}, Ecto.LogEntry]

config :my_app, MyApp.OtherRepo,
  loggers: [{Appsignal.Ecto, :log, [:ecto_other_repo]}, Ecto.LogEntry]
```

It would be nice to update the docs accordingly after this gets released.

### DIFF
--- a/lib/appsignal/ecto.ex
+++ b/lib/appsignal/ecto.ex
@@ -17,7 +17,7 @@ defmodule Appsignal.Ecto do
 
   @nano_seconds :erlang.convert_time_unit(1, :nano_seconds, :native)
 
-  def log(entry) do
+  def log(entry, category \\ "ecto") do
 
     # See if we have a transaction registered for the current process
     case TransactionRegistry.lookup(self()) do
@@ -28,7 +28,7 @@ defmodule Appsignal.Ecto do
         # record the event
         total_time = (entry.queue_time || 0) + (entry.query_time || 0) + (entry.decode_time || 0)
         duration = trunc(total_time / @nano_seconds)
-        Transaction.record_event(transaction, "query.ecto", "", entry.query, duration, 1)
+        Transaction.record_event(transaction, "query.#{category}", "", entry.query, duration, 1)
     end
     entry
   end


jeffkreeftmeijer:
Hey @kanishkablack, thanks for working on this! I discussed this with the rest of the team, and being able to override the event name per database seems like a great addition. We'll do some testing and we'll look at how we want to name this configuration option, and we'll get back to you.

karolsluszniak:
@jeffkreeftmeijer Cool, thanks for looking into this. I'm already using this addition on production and it's a life saver for our app with 2 repos, each hooked to completely different database with different performance properties etc. Here's how it looks in an event timeline:

timeline-for-multiple-repos

tombruijn:
Hi @karolsluszniak,

We've discussed this with the team and see how it's really useful to store some kind of extra metadata on the event to distinguish between repositories and databases. We don't think this is the best way to provide that option right now as it's not very flexible in storing other kinds of data or multiple values. So I'll close this issue. Hopefully you will be able use this solution with your own fork for now.

We have an issue for this on our internal issue tracker where we will continue this discussion. It will probably require a big change of our data format so we will probably work on that for the next major version release.

karolsluszniak:
@tombruijn :+1: Being able to store and browse events by metadata with multiple values would indeed be way better than trying to fit it all into event name, so now I'm eagerly awaiting your next move on this. Meanwhile I'll use my fork as you suggested.

Issue #289, opened 2018-01-09
CompileError: Plug.Conn.WrapperError.__struct__/0 is undefined


larskrantz:
Happens when a creating a new project and having only `:app_signal`-dep in `mix.exs`, and `:plug` is not present. To reproduce:
```bash
mix new foo_appsignal 
cd foo_appsignal
```
add `{:appsignal, "~> 1.0"}` as dep in `mix.exs`.
``` bash
mix deps.get
...
mix compile
...
==> appsignal
Compiling 35 files (.ex)

== Compilation error in file lib/appsignal/error_handler.ex ==
** (CompileError) lib/appsignal/error_handler.ex:117: Plug.Conn.WrapperError.__struct__/0 is undefined, cannot expand struct Plug.Conn.WrapperError
    lib/appsignal/error_handler.ex:117: (module)
could not compile dependency :appsignal, "mix compile" failed. You can recompile this dependency with "mix deps.compile appsignal", update it with "mix deps.update appsignal" or clean it with "mix deps.clean appsignal"
```
Should be solved if changing `{:plug, ">= 1.1.0", optional: true}` to `{:plug, ">= 1.1.0"}` in `appsignal-elixir/mix.exs`.

jeffkreeftmeijer:
Hey Lars,

Thanks for letting us know about this! This is a regression introduced by #281, by referencing the `Plug.Conn.WrapperError` module without wrapping the clause in a `Appsignal.plug?`-check. We'll add a check around it to make sure the library works without having to require Plug, and without depending on it. We'll patch this up and we'll get it ready in the next release.

In the meantime, adding Plug to your dependencies fixes this issue. Sorry about this, and thanks again for letting us know!

larskrantz:
Thanks!

jeffkreeftmeijer:
Hey @larskrantz,

We've released version 1.4.6 yesterday, which includes the patch for this issue. Could you try upgrading to that by running `mix deps.update appsignal` and removing the Plug dependency to see if the new version works for you without having to depend on Plug? Thanks for reporting this issue!

larskrantz:
Sorry for late reply @jeffkreeftmeijer -- but it works fine compiling now!
Thanks /Lars

PR #290, opened 2018-01-09
Making plug a required dependency, fixing #289

larskrantz:


jeffkreeftmeijer:
Hey Lars,

Thanks for taking the time to dive in. As briefly explained in #289, we'd prefer to fix this by not defining the function clause if Plug is not available. That way, we don't have to add a required dependency on Plug. We'll publish a new release that works without Plug, and we'll let you know when it comes out. In the meantime, feel free to use your fork (or add the Plug dependency to your project).

PR #322, opened 2018-03-28
Relax :httpoison dependency to allow ~> 1.0

### LINKED ISSUE #320: Allow httposion 1.0.0
I keep getting httposion dependency version conflicts because this package dictates `{:httpoison, "~> 0.11"},` while other packages in my project (eg. [google_maps](https://github.com/sntran/ex_maps/blob/d2332400491e1e240f85d99edf5fc80e702d05a7/mix.exs#L46)) dictate: `{:httpoison, "~> 1.0"}`. I can fix this with an `override: true` but I'm starting to have to put that in more and more places. At first glance it looks like [the changes](https://github.com/edgurgel/httpoison/compare/v0.13.0...v1.0.0) to 1.0.0 are not that significant.

Can we change the dependency in this package to allow something like `{:httpoison, "< 2.0"}`?

jeffkreeftmeijer:
Closes #320.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -89,7 +89,7 @@ defmodule Appsignal.Mixfile do
 
   defp deps do
     [
-      {:httpoison, "~> 0.11"},
+      {:httpoison, "~> 0.11 or ~> 1.0"},
       {:poison, ">= 1.3.0"},
       {:decorator, "~> 1.2.3"},
       {:plug, ">= 1.1.0", optional: true},


morganick:
👋 @tombruijn we have an app that depends on httpoison 1.0.0. Thanks for updating this!

alChaCC:
i'm also waiting for this fix otherwise can't use Appsignal

jeffkreeftmeijer:
Hey @alChaCC,

We're working on getting this released soon. In the meantime, could you try using `override: true` in your HTTPoison dependency, to see if that allows you to install the AppSignal package? That should allow using a newer version of HTTPoison, which should work as HTTPoison didn't introduce any breaking changes in its more recent versions.

PR #438, opened 2019-01-04
Get tests to pass on Elixir 1.8

jeffkreeftmeijer:
Aside from the fixes in #437, there were some tests failing on 1.8.0:

1. `System.cwd/0` is deprecated and throws a warning.
2. Had to increase the sleeps in ErrorMatcherTest to get them to work. Removing sleeps in favor of retries in the test suite is on my list.
3. Made a stacktrace regex more flexible, because code got moved around in 1.8

### DIFF
--- a/test/appsignal/error_handler/error_matcher_test.exs
+++ b/test/appsignal/error_handler/error_matcher_test.exs
@@ -191,7 +191,7 @@ defmodule Appsignal.ErrorHandler.ErrorMatcherTest do
 
     assert_stacktrace(stacktrace, [
       ~r{\(elixir\) lib/float.ex:\d+: Float.ceil/2},
-      ~r{\(elixir\) lib/task/supervised.ex:\d+: Task.Supervised.do_apply/2},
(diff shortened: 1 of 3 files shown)


test/appsignal/error_handler/error_matcher_test.exs
```diff
@@ -153,7 +153,7 @@ defmodule Appsignal.ErrorHandler.ErrorMatcherTest do
   test "Crashing GenServer with function error", %{fake_transaction: fake_transaction} do
     CrashingGenServer.start(:function_error)
 
-    :timer.sleep(20)
+    :timer.sleep(100)
```
jvanbaarsen: Is this an arbitrary value? Or actually linked to a GenServer timeout value?

jeffkreeftmeijer: This is an arbitrary value that works on Travis. We should look into retrying this a couple of times to decrease the sleep time here, someday.

PR #443, opened 2019-01-12
Swap Poison to Jason

asummers:
Also ABCd the deps list in `mix.exs` while I was there.


jeffkreeftmeijer:
Hey @asummers,

Thanks for raising this issue, and thanks for taking the time to write a patch for this once again. We'd like to switch to using Jason over Poison, as it's now the default in Phoenix. However, we'd like to take this opportunity to make sure we're flexible in regards to which JSON encoder is used, if at all possible.

Ideally, we'd allow users to drop back to Poison if that's the only JSON encoder installed. We'll do some testing to see if that's something we can do, and will get back to you on this. In the meantime, if you (or somebody else reading along) have any ideas on how to do this, please let us know!

asummers:
Oh my fault I had forgotten that we had that discussion already. I deal with a lot of projects.

jeffkreeftmeijer:
We've decided to allow both Poison and Jason as JSON encoders. We'll do so by making both dependencies optional, and having the user depend on one in their projects either explicitly, or via another dependency. We'll check which encoder is available internally. We'll crash the diagnose and install commands if no JSON encoder is available with a message explaining to add one to the project's dependencies.

We'll continue work on this pull request. Thanks for your help so far, @asummers!

jvanbaarsen:
Hi @asummers,

Thanks again for your contribution, we really appreciate it. As @jeffkreeftmeijer mentioned above we want to solve this in a more structured way where the user can pick between Jason and Poison, depending on what they already use in their project.

For this we will create a new PR. I'll close this one for now.

PR #447, opened 2019-01-18
Handle transaction lookup from child processes

bartekbsh:
I have encountered problems when monitoring Ecto, where not all of the queries were reported to the Appsignal. Ecto uses parallel tasks spawned with `Task.async_stream/4` to perform queries when preloading associations outside of a transaction.

Implementation is based on Elixir 1.8 feature. Whenever a process is spawned using `Task` module, newly created process dictionary has `:"$callers"` entry populated, which lists owner pid. If the transaction is not found in the registry, caller pids are used for lookup.


jeffkreeftmeijer:
Hey @bartekbsh! 👋

Thanks for picking this up so quickly, and thanks for making sure to add a test and keeping your patch match the existing API, without breaking on earlier Elixir versions. We’ve worked on ways to get this to work in the past, and we were excited to see `$callers` being added in 1.8 to allow automatic instrumentation for spawned Tasks.

I’ve done some testing in a Phoenix app over multiple Elixir versions, and this patch looks solid so far. For the other reviewers, I’ve tried wrapping an Ecto query in a Task, like this.

```
  def index(conn, _params) do
    task =
      Task.async(fn ->
        Accounts.list_users()
      end)

    users = Task.await(task) |> IO.inspect()
    render(conn, "index.html", users: users)
  end
```

That query is automatically added to the Transaction on 1.8, and it degrades gracefully on lower versions.

Thanks again, Bartek. We’ll do some more testing before merging this, but we’re very excited to ship this in a future version of our integration.

epilgrim:
i deployed something like this in my ecto logger on production, and while it works on web requests without any issue, it brought me another problem.
I also monitor mix tasks. The number of queries in the main process of the mix tasks is minimal, and they get monitored properly. As soon as we start tracking things that happen in related processes, the monitoring is extremely unreliable. I suspect that because of the number of queries i was monitoring was big, the agent started dropping executions of the tasks
I ended up disabling the monitoring of "$callers" for mix tasks, with a configuration flag.

jeffkreeftmeijer:
Thanks for chiming in @epilgrim! You touched on an issue we’ve been having, which is currently preventing us from merging this pull request.

Due to legacy reasons, our agent expects each event (like an Ecto query or a Phoenix controller call) to be run sequentially. It works like a stack, where each started event gets pushed onto the stack, and each finished event pops off again.

This causes issues with multi-process event handling, as sending multiple concurrent events can result in unexpected behaviour. While we allow multi-process requests by passing the Transaction reference to spawned processes we require users to await the processes, and not spawn thousands at once.

However, if we use the `$callers` feature to hook this up automatically, we’re more likely to run into this limitation. Instead of merging this, we’re planning on revisiting multi-process event tracking first, and we’ll come back to this pull request when that’s done.

jeffkreeftmeijer:
We’ve done some research in allowing multi-process event handling in our current agent setup, but we’ve decided to drop that project until we revamp our event handling. I'm closing this for now,  but we’ll be sure to get back to this issue then.

Thanks again for your work on this!

PR #480, opened 2019-04-19
 Implicitly check ignored transactions in `TransactionRegistry.lookup/1`

jeffkreeftmeijer:
Part of #476. 

This patch removes and deprecates `TransactionRegistry.ignored?/0`, which is now handled using `lookup/0` directly instead, to reduce the amount of work being done in the Registry process.

### DIFF
--- a/lib/appsignal/transaction/registry.ex
+++ b/lib/appsignal/transaction/registry.ex
@@ -116,6 +116,7 @@ defmodule Appsignal.TransactionRegistry do
   @doc """
   Check if a progress is ignored.
   """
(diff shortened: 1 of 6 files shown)


lib/appsignal/transaction/registry.ex
```diff
@@ -124,6 +116,7 @@ defmodule Appsignal.TransactionRegistry do
   @doc """
   Check if a progress is ignored.
   """
+  @deprecated "Use Appsignal.TransactionRegistry.lookup/1 instead."
```
jvanbaarsen: Do we keep a changelog for this kind of changes somewhere? (Not sure if customers are actually using this?)

PR #486, opened 2019-04-30
Handle :ignored Transactions in Transaction and Instrumentation.Helpers

jeffkreeftmeijer:
Since #480, `TransactionRegistry.lookup/1` can return `:ignored`. This patch makes sure every function that handles Transactions have a wildcard to ignore those instead of explicitly matching on nil.

### DIFF
--- a/scripts/run_formatter
+++ b/scripts/run_formatter
@@ -4,6 +4,6 @@ version() {
   echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
 }
 
-if [ $(version $TRAVIS_ELIXIR_VERSION) -ge $(version "1.6.6") ]; then
+if [ $(version $TRAVIS_ELIXIR_VERSION) -ge $(version "1.8.0") ]; then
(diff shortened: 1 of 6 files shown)


scripts/run_formatter
```diff
@@ -4,6 +4,6 @@ version() {
   echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
 }
 
-if [ $(version $TRAVIS_ELIXIR_VERSION) -ge $(version "1.6.6") ]; then
+if [ $(version $TRAVIS_ELIXIR_VERSION) -ge $(version "1.8.0") ]; then
```
jvanbaarsen: I believe we don't run these files anymore right? Everything got moved to `scripts/test`

jeffkreeftmeijer: That's right. This is a quick fix on this hotfix to fix a test failure on CI, so we don't have to base on develop right now. This code will be gone once we release a new minor version, as that'll have develop merged in.

jvanbaarsen: ah ok :D

PR #491, opened 2019-05-13
Use the bundled certificate and ciphers when downloading agent

jeffkreeftmeijer:
Like #489, but for downloading the agent. The code is duplicated from
Appsignal.Transmitter, as that module is not yet compiled at compile
time.

### DIFF
--- a/mix_helpers.exs
+++ b/mix_helpers.exs
@@ -187,8 +187,7 @@ defmodule Mix.Appsignal.Helper do
   end
 
   defp download_options do
-    ca_file_path = Path.join(:code.priv_dir(:appsignal), "cacert.pem")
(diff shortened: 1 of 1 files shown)


jeffkreeftmeijer:
That newly added module wouldn’t be available at compile time, unless we do one of two things:

1. Move it to an `.exs` file, making it interpreted code. That’s what we do for `agent.exs`, which we need in compile time. This would require us to evaluate the file file before using any of the functions available in it.
2. Move it to a different OTP app, which the appsignal application depends on. We’ve debated doing this in the past for other things in the mix.exs file, like the code that checks the system’s architecture. Having a separate wrapper for our HTTP client could make sense (appsignal_http?), for example.

Option 2 would be nicest, but we'll most likely have to turn the integration into an umbrella to do that while keeping all of the code contained in one library.

PR #503, opened 2019-08-06
remove handle_call for monitoring and removing a process

BenMorganIO:
This GenServer.call does not obtain information from state or modify it. To increase the throughput of this GenServer we remove a blocking handle_call to reduce unneeded backpressure.


lib/appsignal/transaction/registry.ex
```diff
@@ -36,11 +36,9 @@ defmodule Appsignal.TransactionRegistry do
     pid = self()
 
     if Appsignal.Config.active?() && registry_alive?() do
-      monitor_reference = GenServer.call(__MODULE__, {:monitor, pid})
+      monitor_reference = Process.monitor(pid)
```
BenMorganIO: Now, this may have to go into a GenServer since it may not be linking the proper process.

BenMorganIO:
See PR #504 for benchmark on this PR. I don't recommend merging yet.

PR #504, opened 2019-08-07
Improving Transaction Registry Performance

BenMorganIO:
This PR is all about improving the transaction registry. Currently, since it's a genserver, we end up getting backpressure. This PR is about trying to reduce that backpressure as intelligently as possible.

First up, we got a benchmark. I used the Fibonacci sequence to make a list of concurrent processes and then started hitting it. My initial benchmark for the system:

```
Name               ips        average  deviation         median         99th %
registry         0.109         9.15 s     ±0.00%         9.15 s         9.15 s
```

Then I checked in on what effect PR #503 had on the benchmark:

```
Name               ips        average  deviation         median         99th %
registry        0.0609        16.42 s     ±0.00%        16.42 s        16.42 s
```

Now, this is caused by silent backpressure, so to remove that we alter a `handle_cast` to a `handle_call` and we get:

```
Name               ips        average  deviation         median         99th %
registry         0.113         8.87 s     ±0.00%         8.87 s         8.87 s
```

Looking better, but barely. The problem is now with the GenServer in the transaction registry being the root cause of the backpressure.

The first step was moving ETS to its own file and its own process. We start the process up with an Agent to boot ETS and to make sure that ETS calls to its own process for management if needed in the future (such as changing ownership).

After that, the `init` call was free of responsibilities and we could start moving code over to a task; that's the `Receiver`. We start a simple task, register it, and listen and loop over messages send to the process. This will be the only blocking part.

Then we add two public functions called `monitor/1` and `demonitor/1` for the `Receiver` to monitor and demonitor. We update the code in the `TransactionRegistry` and benchmark:

```
Name               ips        average  deviation         median         99th %
registry         0.122         8.19 s     ±0.00%         8.19 s         8.19 s
```

The **average duration is 10.5% smaller** than the initial, the **ips is 10.6% larger**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ you have [Elixir installed](http://elixir-lang.org/install.html).
 Then make sure you have all the project's dependencies installed by running the
 following command:
 
-```
+```shell
 mix deps.get
 ```
 
@@ -62,7 +62,7 @@ Testing is done with ExUnit and can be run with the `mix test` command. You can
 also supply a path to a specific file path you want to test and even a specific
 line on which the test you want to run is defined.
 
-```
+```shell
 mix test
 mix test test/appsignal/some_test.ex:123
 ```
@@ -71,7 +71,7 @@ This project has several different test suites defined with different mix
 environments. You can run them by specifying the specific type of test suite in
 the `MIX_ENV` environment variable.
 
-```
+```shell
 # Default
 MIX_ENV=test mix test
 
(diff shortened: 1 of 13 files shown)


BenMorganIO:
@tombruijn I would recommend relabelling this to critical performance fix ;)

jeffkreeftmeijer:
Hey @BenMorganIO,

Thanks for taking the time to dive into this, and thanks for submitting such a well-polished pull request!

 We’re looking into this, but in order not to break any existing installs, we’ll need to do some thorough testing to make sure everything still functions as before to be able to merge this. We’ll keep you posted on our progress with this.

In the meantime, please feel free to run your fork, and please let us know if you run into any issues.

BenMorganIO:
@ jeffkreeftmeijer FWIW, we haven't seen any new appsignal timeouts in our logs.

jeffkreeftmeijer:
Hey @BenMorganIO, after merging some of your changes into develop already, I've opened a new pull request with the rest of your changes in #505 to get reviewed by the rest of the team. I'm closing this one, let's continue there.

PR #521, opened 2019-10-03
Don't match on return from `Transaction.complete/1`

jeffkreeftmeijer:
Since 1.11.0, `Appsignal.Transaction.complete/1` can return `nil` if
there was no Transaction to be completed. When AppSignal is disabled and
while using transaction decorators, this could result in a broken
match. This patch removes the match.


EskiMag:
It would be more convenient to have this change marked as a *breaking change* or something similar in the CHANGELOG. :-) Without that I had to investigate for a few minutes. No big deal, but just an improvement to consider.

jeffkreeftmeijer:
Hey @EskiMag,

Good call! We'll keep that in mind in the future. :)

PR #527, opened 2019-11-19
Add explicit error handling for Phoenix channels

jeffkreeftmeijer:
In current versions of the integration, exceptions raised in channel actions fall through and get picked up by the error handler, which adds the exception to the Transaction that's opened by the channel instrumentation.

This, however, skips setting the environment data, which happens at the end of the Transaction for performance samples. The code never reaches that, as the exception isn't caught.

This patch adds a `try`-block to the channel instrumentation to make sure the error is caught there, and allowing the environment data to be set in the `catch` block.

### DIFF
--- a/lib/appsignal/phoenix/channel.ex
+++ b/lib/appsignal/phoenix/channel.ex
@@ -95,19 +95,37 @@ if Appsignal.phoenix?() do
       <<"Elixir.", action::binary>> = action_str
       @transaction.set_action(transaction, action)
 
-      result = function.()
+      try do
+        function.()
+      catch
+        kind, reason ->
+          stack = System.stacktrace()
 
-      resp = @transaction.finish(transaction)
+          {exception, stacktrace} = Appsignal.Error.normalize(reason, stack)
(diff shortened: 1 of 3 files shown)


lib/appsignal/phoenix/channel.ex
```diff
@@ -94,11 +94,32 @@ if Appsignal.phoenix?() do
       <<"Elixir.", action::binary>> = action_str
       @transaction.set_action(transaction, action)
 
-      result = function.()
-
-      resp = @transaction.finish(transaction)
+      try do
+        function.()
+      catch
```
xadamy: This is a fairly large block with no comments. Can we get a comment or two explaining what the intent is here?

jeffkreeftmeijer: You're right, this block was duplicating code from the ErrorHandler, because that implementation takes care of closing the Transaction, and expects a `%Plug.Conn{}` (which we don't have for channels).

I've created a new function in the ErrorHandler called `set_error/3`, which doesn't take a conn and just sets the error, so we move most of the error unpacking logic there.

PR #528, opened 2019-11-22
Add Phoenix Telemetry event handler to add Phoenix 1.5 instrumentation

jeffkreeftmeijer:
Phoenix 1.5 [removes the instrumentation API](https://github.com/phoenixframework/phoenix/commit/19b0b01e2fc751901e06fcd74db1fc5b39672d26) in favour of using Telemetry. This pull request adds a Telemetry event handler that catches `[:phoenix, :endpoint, :start]` and `[:phoenix, :endpoint, :stop]` events to create an event named `call.endpoint`, which is comparable to `call.controller` in the current version.

> Before version Phoenix version 1.5. AppSignal’s Phoenix instrumentation depended on data from the Phoenix instrumenter, and the installation instructions included a step to attach AppSignal’s instrumenter to your application in your app’s configuration:
> 
> ```
> config :appsignal_phoenix_example, AppsignalPhoenixExampleWeb.Endpoint,
>   #...
>   instrumenters: [Appsignal.Phoenix.Instrumenter]
> ```
> 
> From Phoenix 1.5 on, the old Phoenix instrumentation is deprecated and removed in favor of the new Telemetry-based instrumentation. When upgrading to Phoenix 1.5, you’ll see a warning during compilation when using the old instrumenters:
> 
> ```
> [warn] :instrumenters configuration for AppsignalPhoenixExampleWeb.Endpoint is deprecated and has no effect
> ```
> 
> To switch to the new instrumentation, make sure you’re running version 1.12.0 of the AppSignal integration or higher. Then, remove the instrumenters configuration option from your endpoint configuration.
> 
> The new instrumentation should appear automatically in your samples as an event named `call.phoenix_endpoint`. To gain more insights into your app, you can.

# Reviewing

To review this patch, remove the instrumenter from your configuration like described above, and check to make sure you're still getting endpoint events in AppSignal when you request a controller action. Also, feel free to un-request your review, I've requested reviews for all of you make sure I could find someone with time to test this.

# TODO:
- [x] Take the current Transaction from the conn, instead of doing a lookup for both events
- [x] Set the action name from the endpoint start event
- [x] Make sure the current implementation works with multiple Phoenix apps in an umbrella
- [ ] Determine if the automatic setup (through `EventHandler.attach/0`) works for different application setups

PR #538, opened 2020-01-16
Set Phoenix 1.5 action names in the EventHandler

jeffkreeftmeijer:
After #528, which adds Telemetry-based instrumentation to add support for Phoenix 1.5—which removes the `Phoenix.Instrumenter` module we previously relied on—this patch adds an event handler for `[:phoenix, :router_dispatch, :start]` events, which include the controller and action names in their metadata.

Since this event is called before the action is executed, we can set the action name directly on the Nif, and have it be overwritten in the controller action by calling the same function again later. Also, unwrapped errors (like exits) have their action names properly added if they happen in a controller action.

To test this, use a Phoenix 1.5 application and make sure the action names are set correctly and can be overwitten. Also, make sure errors have the correct action names added. A good app to test with is [appsignal_phoenix_example, in the phoenix-1.5 branch](https://github.com/jeffkreeftmeijer/appsignal-phoenix-example/tree/phoenix-1.5), which is updated to Phoenix 1.5 and depends on this branch of AppSignal for Elixir.

### DIFF
--- a/lib/appsignal/phoenix/event_handler.ex
+++ b/lib/appsignal/phoenix/event_handler.ex
@@ -1,4 +1,6 @@
 defmodule Appsignal.Phoenix.EventHandler do
+  import Appsignal.Utils
+
   @transaction Application.get_env(:appsignal, :appsignal_transaction, Appsignal.Transaction)
 
   @spec attach() :: :ok | {:error, :already_exists}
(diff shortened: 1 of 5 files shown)


jeffkreeftmeijer:
Removed reviewers and moved this back to in progess to try an [approach that's hopefully easier recommended by José](https://elixirforum.com/t/replacing-phoenix-instrumenter-with-telemetry-events-in-appsignals-phoenix-integration/28325/2?u=jeff).

Issue #546, opened 2020-03-09
Logging conflict with Plugs that take modules as options after v1.12


bgentry:
## Describe the bug

After upgrading to absinthe-elixir v1.12, we began to see these errors logged in our tests:

```
[error] Handler "appsignal_phoenix_event_handler" has failed and has been detached. Class=:error
Reason=%ArgumentError{message: "cannot convert the given list to a string.\n\nTo be converted to a string, a list must either be empty or only\ncontain the following elements:\n\n  * strings\n  * integers representing Unicode code points\n  * a list containing one of these three elements\n\nPlease check the given list or call inspect/1 to get the list representation, got:\n\n[schema: MyApp.Web.GraphQL.Schema]\n"}
[...]

PR #547, opened 2020-03-10
EventHandler handles router_dispatch events with plug_opts

jeffkreeftmeijer:
As reported in #546, mounted plugs have the ability to pass non-atom
plug_opts. This patch makes sure the Phoenix.EventHandler ignores them
instead of trying to convert the options to an action name.

### DIFF
--- a/lib/appsignal/phoenix/event_handler.ex
+++ b/lib/appsignal/phoenix/event_handler.ex
@@ -25,8 +25,13 @@ defmodule Appsignal.Phoenix.EventHandler do
         _measurements,
         %{plug: controller, plug_opts: action},
         _config
-      ) do
+      )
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
To help test this patch, I've added a [plug-with-opts](https://github.com/jeffkreeftmeijer/appsignal-phoenix-example/tree/plug-with-opts) branch to the Phoenix example app. Requesting `/plug` shouldn't crash the event handler and shouldn't print an error to the log.

jeffkreeftmeijer:
⚠️  Please don't delete this branch, as users have switched their dependencies to it (#546). All users running the branch from git are urged to upgrade to 1.12.1.

PR #548, opened 2020-03-11
Add LiveView module

isavita:
That is my attempt for adding basic functionality allowing tracking `LiveView`.
Is there any plans for adding such a functionally?
If there is I can continue working on this PR.

1. Adding test.
2. Adding proper versioning for which version of Phoenix is required and which version of live_view.
3. Improve the module API by supporting more functionality, better naming.

Any thoughts?

PR #549, opened 2020-03-12
Add LiveView support

jeffkreeftmeijer:
Based on @isavita's pull request in #548, this pull request continues work to finish the work on adding LiveView support, and is drafted to get CI to run tests.

### DIFF
--- a/lib/appsignal.ex
+++ b/lib/appsignal.ex
@@ -64,6 +64,10 @@ defmodule Appsignal do
     Code.ensure_loaded?(Phoenix)
   end
 
+  def live_view? do
+    phoenix?() && Code.ensure_loaded?(Phoenix.LiveView)
+  end
+
   @doc """
   Application callback function
   """
--- a/lib/appsignal/instrumentation/decorators.ex
+++ b/lib/appsignal/instrumentation/decorators.ex
@@ -34,7 +34,9 @@ defmodule Appsignal.Instrumentation.Decorators do
     transaction: 1,
     transaction_event: 0,
     transaction_event: 1,
-    channel_action: 0
+    channel_action: 0,
+    live_view_mount: 0,
+    live_view_event: 0
 
   import Appsignal.Utils
 
@@ -84,6 +86,32 @@ defmodule Appsignal.Instrumentation.Decorators do
         unquote(module_name(context.module)),
         unquote(action),
         unquote(socket),
+        fn -> unquote(body) end
+      )
+    end
+  end
+
+  @doc false
+  def live_view_mount(body, %{args: [_session, payload, socket]} = context) do
+    quote do
+      Appsignal.Phoenix.LiveView.live_view_action(
+        unquote(module_name(context.module)),
+        unquote(context.name),
+        unquote(socket),
+        unquote(payload),
+        fn -> unquote(body) end
+      )
+    end
+  end
+
+  @doc false
+  def live_view_event(body, %{args: [event, payload, socket]} = context) do
+    quote do
+      Appsignal.Phoenix.LiveView.live_view_action(
+        unquote(module_name(context.module)),
+        unquote("#{context.name}(\"#{event}\")"),
+        unquote(socket),
+        unquote(payload),
         fn -> unquote(body) end
       )
     end
--- a/lib/appsignal/instrumentation/decorators.ex
+++ b/lib/appsignal/instrumentation/decorators.ex
@@ -36,7 +36,8 @@ defmodule Appsignal.Instrumentation.Decorators do
     transaction_event: 1,
     channel_action: 0,
     live_view_mount: 0,
-    live_view_event: 0
+    live_view_event: 0,
+    live_view_handle_params: 0
 
   import Appsignal.Utils
 
@@ -92,7 +93,7 @@ defmodule Appsignal.Instrumentation.Decorators do
   end
 
   @doc false
-  def live_view_mount(body, %{args: [_session, payload, socket]} = context) do
+  def live_view_mount(body, %{args: [_params, payload, socket]} = context) do
     quote do
       Appsignal.Phoenix.LiveView.live_view_action(
         unquote(module_name(context.module)),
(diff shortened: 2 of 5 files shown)


jeffkreeftmeijer:
Hey @isavita,

Let's continue here, as I've done some work to get this ready to be merged. I've made sure we don't add the LiveView dependency to apps running on an Elixir version below 1.7, and I've added regression tests for the `Appsignal.Phoenix.LiveView` module. 

However, I'm unsure about the decorators. Since LiveView events can be called to in multiple ways, it seems like it'd be difficult to have decorators that work for all cases. For example, using `send` and `handle_info` with the live_view_event decorator currently breaks as there's no function clause without parameters.

```
@decorate live_view_event
def handle_info(:tick, socket) do
```

Also, in the current implementation, the event name extraction fails if it's not an atom;

```
@decorate live_view_event
def handle_info(name, socket) do
```

Instead of trying to get this perfect, I'm thinking of dropping the decorators and only adding the instrumentation helpers in Appsignal.Phoenix.LiveView, to be used like our [channel instrumentation helpers](https://docs.appsignal.com/elixir/instrumentation/instrumentation.html#instrumentation-helper-functions).

isavita:
Thank you @jeffkreeftmeijer !
The `live_view_event` was for [handle_event](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_event/3). It is always binary. Probably better name will be `live_view_handle_event`.
I agree for the decorators for that when in the path to have them explicit, but name this one badly.
The [handle_info](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_info/2) can be `term`. I would imagine to add `live_view_handle_info` for it. 

Live view has 5 [callbacks](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#callbacks).

I was imagining having explicit decorator for each.
```elixir
@decorate mount
def mount(:tick, socket) do
```
```elixir
@decorate live_view_handle_event
def handle_event(:tick, socket) do
```
```elixir
@decorate live_view_handle_info
def handle_info(:tick, socket) do
```

jeffkreeftmeijer:
Hey @isavita,

Thanks once again for your work on this patch. We've just merged LiveView support into develop and released 1.13.0-beta.1 with the instrumentation helpers. We'll look into adding decorator-like helpers in a future version of our integration, but we wanted to ship this to our users so they can start instrumenting their live views, and gain some feedback on the implementation.

For everyone reading along; please try the beta and let us know how the helpers work for you. We'll work on adding documentation and will release 1.13.0 after some more testing on our side.

Thanks!

sergiotapia:
Where did this end up @jeffkreeftmeijer ? I don't see any of the liveview decorators on latest. https://github.com/appsignal/appsignal-elixir/blob/24fce5d05f1381f83726d3d5edbfac8ba5fca0d2/lib/appsignal/instrumentation/decorators.ex

Appreciate the help

tombruijn:
Hi @sergiotapia, we moved the live view integration to the [appsignal_phoenix](https://hex.pm/packages/appsignal_phoenix) package, in version 2.0 of the appsignal-elixir package.

sergiotapia:
Thanks @tombruijn my code works and looks like this:

```
@impl true
  def mount(_params, _session, socket) do
    live_view_action(__MODULE__, "mount", socket, fn ->
      changeset = Leads.change_waitlist(%Waitlist{})

      socket =
        assign(socket,
          changeset: changeset,
          waitlist_submitted: get_connect_params(socket)["waitlist_submitted"]
        )

      {:ok, socket, layout: {GamingWeb.LayoutView, "live_no_navbar.html"}}
    end)
  end
```

Is it left as an exercise for the reader to implement this as a tidier decorator?

tombruijn:
We don't have decorators for this part of Phoenix unfortunately. We haven't found a better way to support it yet. We actually want to remove decorators entirely and replace it with something better, like macros. You can write your own macro for this as well. Please share yours if you write one.

sergiotapia:
@tombruijn Thanks Tom, a better way to do this was proposed by Ben Wilson - use the new Liveview hooks. https://sergiotapia.com/integrating-appsignal-with-phoenix-liveview-tutorial

tombruijn:
@sergiotapia that could be interesting! Thanks for writing the article!
I've created an issue for it here: https://github.com/appsignal/appsignal-elixir-phoenix/issues/37

cblavier:
I'm using Phoenix LiveView telemetry events to instrument events & report errors with metadata: https://gist.github.com/cblavier/faaee0e3969de708737ed0e6c623d3dd

The issue now is that I get duplicate errors since they are also handled by the generic background reporter. Any idea how to handle this issue?

cblavier:
@jeffkreeftmeijer if you have any idea how to tackle this issue (duplicate errors mentioned in my last post), I would be happy to open a PR on the appsignal-elixir-phoenix project

jeffkreeftmeijer:
Hey @cblavier,

Something we do in the Plug instrumentation is ignoring the current pid when the error is caught there, so it’s ignored in the error handler using `Appsignal.Tracer.ignore/0`. Would something like that work for your use case?

https://github.com/appsignal/appsignal-elixir-plug/blob/263441640bc06d02c32843a289441ef570f4052c/lib/appsignal_plug.ex#L52-L62

cblavier:
The Telemetry genserver I wrote is actually running in a different process so if I call `Tracer.ignore/0` I guess it won't ignore the correct pid (liveview's pid)

The other issue is that I'm not sure when I will receive the telemetry event (sometime before, sometime after the exception is actually catched by the liveview)

Is there a way to make appsignal merge bug reports based on some unique identifier? 
Or to get the same `span` used if I create a span with a specific pid?

cblavier:
Ok I may have found something working, by using a new `Tracer.ignore/1` function that I use to ignore liveview's pid from my telemetry listener.
WDYT?

cblavier:
The telemetry integration is now running perfectly for us in production. It's really valuable because of the extra context it contains when reporting bugs in LiveView callbacks.

Do you think it could be integrated in your codebase? 
If you give me some guidelines, I can open a PR

cblavier:
I just noticed that using previous gist (telemetry server), I don't have a detailed breakdown of the LiveView event:

CleanShot 2022-02-22 at 10 20 49

Any clue of how I could improve this?

cblavier:
Ok I figured it out and I now have detailed breakdown of LiveviewEvents. I updated my [gist](https://gist.github.com/cblavier/faaee0e3969de708737ed0e6c623d3dd)
@jeffkreeftmeijer do you think it deserves a PR?

jeffkreeftmeijer:
Hey @cblavier,

Thanks for working on this! We have [a pull request open for auto-instrumenting LiveView events](https://github.com/appsignal/appsignal-elixir-phoenix/pull/40) right now, so we’re looking into that as well. The question now is how we’d want to support LiveView going forward. If we can do what we’re doing with LiveView now through `:telemetry` completely, that’d be my preference right now. However, I haven’t had time to properly dive into this. 

If it’s not too much work, could you throw your code in a pull request on appsignal-elixir-phoenix to continue the conversation there?

PR #555, opened 2020-04-14
554 add record event callback

### LINKED ISSUE #554: TransactionBehaviour doesn't include the `record_event` callback
## Describe the bug

We're using `Appsignal.TransactionBehaviour` to be able to use [mox](https://github.com/dashbitco/mox) to test our telemetry handlers that report to appsignal. However, we use `record_event` quite a bit, but we're not able to mock it easily because the behaviour doesn't implement a callback for this function.

## To Reproduce

- Using AppSignal for Elixir package version 1.12.1
- With this code:
  ```elixir
  Mox.defmock(Sketchql.Monitoring.Appsignal.TransactionMock, for: Appsignal.TransactionBehaviour)
  ```
When testing something that calls `Appsignal.Transaction.record_event/5`, you'll see the following error:

```
** (ArgumentError) unknown function record_event/4 for mock TransactionMock
```

uesteibar:
Fixes #554 

Additionally, I also added the callbacks for `set_namespace` as they were also missing.

### DIFF
--- a/lib/appsignal/transaction.ex
+++ b/lib/appsignal/transaction.ex
@@ -14,6 +14,16 @@ defmodule Appsignal.TransactionBehaviour do
               Appsignal.Transaction.t() | nil
   @callback set_sample_data(Appsignal.Transaction.t() | any(), String.t(), any) ::
               Appsignal.Transaction.t() | nil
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Hey @uesteibar, thanks a lot for picking this up!

PR #559, opened 2020-04-29
Use `__STACKTRACE__ /0` on Elixir >= 1.7

jeffkreeftmeijer:
`System.stacktrace/0` was deprecated and replaced with `__STACKTRACE__/0` in Elixir 1.7. Current versions throw deprecation warnings, so we're moving to using the new method on versions that have it available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ branches:
   only:
   - master
   - develop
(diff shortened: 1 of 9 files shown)


jeffkreeftmeijer:
Squash-merged!

PR #568, opened 2020-06-09
run Erlang/OTP v23.0 in Travis CI

bgentry:
Erlang/OTP v23 should work properly now that Hackney v1.16.0 is released. However, we weren't running this version in CI up to this point, so we can't be sure. This PR adds OTP v23 to the build matrix so it will run from now on.

While I was at it, I also bumped the latest Elixir version to v1.10.3. I didn't opt to remove any old OTP versions, don't want to assume whether you'd like to drop support for any older versions.


bgentry:
Actually, it looks like this won't work yet—Hex doesn't yet host Elixir builds for OTP 23:

```
Installing Elixir 1.10.2
$ wget https://repo.hex.pm/builds/elixir/v1.10.2-otp-23.zip
--2020-06-09 00:37:12--  https://repo.hex.pm/builds/elixir/v1.10.2-otp-23.zip
Resolving repo.hex.pm (repo.hex.pm)... 151.101.2.2, 151.101.66.2, 151.101.130.2, ...
Connecting to repo.hex.pm (repo.hex.pm)|151.101.2.2|:443... connected.
HTTP request sent, awaiting response... 403 Forbidden
2020-06-09 00:37:12 ERROR 403: Forbidden.
```

That happens for both Elixir v1.10.2 and v1.10.3.

jeffkreeftmeijer:
Hey @bgentry,

Thanks for picking this up. We'll keep an eye out so we can merge this when the builds are available.

PR #611, opened 2020-12-24
Let set error use root span

thijsc:
Set error used the current span, which caused errors to not be tracked. Always use the root span instead.


jeffkreeftmeijer (approved):
Added a regression test to make sure we use the root span even if a child span exists.

xadamy (approved):

tombruijn (approved):

PR #612, opened 2020-12-30
Fixed some warnings

oliver-kriska:
Fixed some small warnings

### DIFF
--- a/lib/appsignal/ecto.ex
+++ b/lib/appsignal/ecto.ex
@@ -55,9 +55,11 @@ defmodule Appsignal.Ecto do
     handle_query(@tracer.current_span(), total_time, repo, query)
   end
 
+  def handle_event(_event, _measurements, _metadata, _config), do: :ok
+
   defp handle_query(nil, _total_time, _repo, _query), do: nil
 
-  defp handle_query(current, total_time, repo, query) do
+  defp handle_query(_current, total_time, repo, query) do
     time = :os.system_time()
 
     "http_request"
@@ -67,6 +69,4 @@ defmodule Appsignal.Ecto do
     |> @span.set_sql(query)
     |> @tracer.close_span(end_time: time)
   end
-
-  def handle_event(_event, _measurements, _metadata, _config), do: :ok
 end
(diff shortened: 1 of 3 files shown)


jeffkreeftmeijer:
Awesome, thanks @oliver-kriska! This’ll be in the next release. 👍

oliver-kriska:
To be clear I fixed those:
```
warning: variable "pid" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/appsignal/transaction_registry.ex:5: Appsignal.TransactionRegistry.lookup/1

warning: variable "parent" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/appsignal/transaction_registry.ex:9: Appsignal.TransactionRegistry.register/1

warning: module attribute @deprecated in code block has no effect as it is never returned (remove the attribute or assign it to _ to avoid warnings)
  lib/appsignal/transaction.ex:7: Appsignal.Transaction

warning: undefined module attribute @deprecated, please remove access to @deprecated or explicitly set it before access
  lib/appsignal/transaction.ex:7: Appsignal.Transaction (module)

warning: variable "current" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/appsignal/ecto.ex:60: Appsignal.Ecto.handle_query/4

warning: clauses with the same name and arity (number of arguments) should be grouped together, "def handle_event/4" was previously defined (lib/appsignal/ecto.ex:51)
  lib/appsignal/ecto.ex:71
```

I these are not possible to fix right now:
```

warning: Supervisor.Spec.worker/2 is deprecated. Use the new child specifications outlined in the Supervisor module instead
Found at 3 locations:
  lib/appsignal.ex:29: Appsignal.start/2
  lib/appsignal.ex:30: Appsignal.start/2
  lib/appsignal.ex:31: Appsignal.start/2

warning: Appsignal.Instrumentation.instrument/4 is deprecated. Use Appsignal.instrument/3 instead.
  lib/appsignal/instrumentation/helpers.ex:6: Appsignal.Instrumentation.Helpers.instrument/4
  ```
First one would be drop support for older Elixir and the last one will be fixed when you remove deprecated code.

PR #631, opened 2021-01-27
Use pid from conn in Error.Backend if available

jeffkreeftmeijer:
Appsignal.Plug [ignores](https://github.com/appsignal/appsignal-elixir-plug/blob/6eda8784eb27365077f3dce6b75983eec6d02b22/lib/appsignal_plug.ex#L61) the current process whenever it catches an exception. That’s done to allow `Appsignal.Error.Backend` to [drop](https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/error/backend.ex#L25-L26) any error reports that came from that process, as those were already handled by `Appsignal.Plug`.

Since plug_cowboy 2.0.2 (https://github.com/elixir-plug/plug_cowboy/commit/6513e38529324283dc7ac430e9f8a3107dbd936d, to be exact), the pid that’s sent to the error logger doesn’t match the pid we’re ignoring. This causes duplicate errors, where one is logged to the “web” namespace (the one caught by `Appsignal.Plug`) and another is logged in the “background” namespace (the one from `Appsignal.Error.Backend`).

Since https://github.com/elixir-plug/plug_cowboy/commit/1f2850086d5f26115878d65dccfd41b2b0ebcaa1, the `conn` is added to the report, allowing us to fetch the pid through the `conn`’s `owner` instead. This patch uses that instead of the `pid` if it’s available. This would still allow us to get errors from the `:cowboy` domain, just not from errors we’ve already caught elsewhere.

~Note: this is only available in a yet-to-be-released version of plug_cowboy.~ No longer true, this patch works without it as well: https://github.com/appsignal/appsignal-elixir/pull/631#issuecomment-768224900

Closes https://github.com/appsignal/support/issues/107.

### DIFF
--- a/lib/appsignal/error/backend.ex
+++ b/lib/appsignal/error/backend.ex
@@ -42,6 +42,12 @@ defmodule Appsignal.Error.Backend do
     end
   end
 
+  defp handle_report(%{crash_reason: {reason, stacktrace}, pid: pid}) do
+    pid
+    |> @tracer.lookup()
+    |> do_handle_report(pid, reason, stacktrace)
+  end
+
   defp do_handle_report([{_pid, :ignore}], _, _, _) do
     :ok
   end
--- a/lib/appsignal/error/backend.ex
+++ b/lib/appsignal/error/backend.ex
@@ -48,6 +48,10 @@ defmodule Appsignal.Error.Backend do
     |> do_handle_report(pid, reason, stacktrace)
   end
 
+  defp handle_report(_) do
+    :ok
+  end
+
   defp do_handle_report([{_pid, :ignore}], _, _, _) do
     :ok
   end
--- a/lib/appsignal/error/backend.ex
+++ b/lib/appsignal/error/backend.ex
@@ -17,33 +17,45 @@ defmodule Appsignal.Error.Backend do
   end
 
   def handle_event({:error, gl, {_, _, _, metadata}}, state) when node(gl) == node() do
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Reading up on https://github.com/elixir-plug/plug_cowboy/issues/55 and looking into Sentry’s implementation I found that they ignore the [`:cowboy`](https://github.com/getsentry/sentry-elixir/blob/121464782e522d59a26ffd523fffb84eb746a357/lib/sentry/logger_backend.ex#L47) domain completely.

This has the advantage of working today, without having to wait for the cowboy_plug patch. Although I’m not totally in favor of dropping everything (I’d prefer to keep the ability to still catch any error that might fall through our Plug wrapping), an extra clause we could add is to still drop the error if there’s no `conn` in the report, but if there’s a `pid` that’s not ignored.

That’d mean we mimic Sentry’s solution for now (which is dropping all reports from cowboy), but we’ll upgrade to dropping only ignored errors when the plug_cowboy version is released.

jeffkreeftmeijer:
Added https://github.com/appsignal/appsignal-elixir/pull/631#issuecomment-768224900 in 741fc1702eda0718a018b1c16887136148b26990.

jeffkreeftmeijer:
This is ready for review, by the way.

thijsc (approved):

tombruijn (approved):

PR #650, opened 2021-05-27
Replacing AppSignal.Monitor's list state with a MapSet

schrockwell:
The `Appsignal.Monitor` GenServer was storing a list of pids, but since uniqueness is guaranteed (you can't monitor a particular pid more than once), MapSet is a better data structure to store this information, since all its operations are constant-time instead of linear-time. 

Considering that this GenServer was previously a performance bottleneck in v2.1.6, I think this incremental improvement makes sense to curb potential future performance issues.

### DIFF
--- a/lib/appsignal/monitor.ex
+++ b/lib/appsignal/monitor.ex
@@ -10,10 +10,10 @@ defmodule Appsignal.Monitor do
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
   end
 
-  def init(state) do
+  def init(_) do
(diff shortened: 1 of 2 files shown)


shairyar:
Thanks @schrockwell for the PR we will take a look.

jeffkreeftmeijer:
Hey @schrockwell,

Thanks a lot for diving in here. Using a `MapSet` definitely makes sense, indeed! I’ll merge this one into develop to release it as part of 2.2.0. Thanks again!

PR #652, opened 2021-06-15
Improve error names

TheFirstAvenger:
Right now, all error events on the Incidents -> Errors -> Issues List page are slotted into a few large error buckets, such as `FunctionClauseError`, `CastError`, `ErlangError`, etc... This is because the "name" is simply the struct name of the Exception module. The problem with this is that you could have dozens of different `FunctionClauseError`s, each needing to be triaged, submitted to you ticketing system, and tracked independently. This is not possible to do currently, as they all show up as different "samples" of the same Error. This PR integrates the beginning of the actual error message in the error name, causing same error structs with different causes to be tracked independently.

### DIFF
--- a/lib/appsignal/error.ex
+++ b/lib/appsignal/error.ex
@@ -1,9 +1,11 @@
 defmodule Appsignal.Error do
   @moduledoc false
   def metadata(%_{__exception__: true} = exception, stack) do
+    banner = Exception.format_banner(:error, exception, stack)
+
     {
-      inspect(exception.__struct__),
-      Exception.format_banner(:error, exception, stack),
+      name(banner, exception.__struct__),
+      banner,
       Appsignal.Stacktrace.format(stack)
     }
   end
@@ -15,10 +17,20 @@ defmodule Appsignal.Error do
   end
 
   def metadata(kind, reason, stack) do
+    banner = Exception.format_banner(kind, reason, stack)
+
     {
-      inspect(kind),
-      Exception.format_banner(kind, reason, stack),
+      name(banner, kind),
+      banner,
       Appsignal.Stacktrace.format(stack)
     }
+  end
+
+  defp name(banner, type) do
+    if String.contains?(banner, inspect(type)) do
+      banner
+    else
+      type <> " " <> banner
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


shairyar:
Hey @TheFirstAvenger thank you for submitting the PR, we will take a look.

shairyar:
Hey @TheFirstAvenger, thanks for the PR! We have decided not to accept this PR, we see this is a problem and we are currently working on better error grouping and intend to finish that project sometime soon. Once we have something released you can test we will contact you so you can give that a try and let us know your feedback.

Could you drop us an email at support@appsignal.com saying that you are interested in testing the beta so we can then inform you?

TheFirstAvenger:
@shairyar what is the current status of the work you mention here on better error grouping? Running the latest as of the end of last year, I am still seeing names like `FunctionClauseError`, `CastError`, `ErlangError`, etc...

shairyar:
@TheFirstAvenger you can enable the error grouping strategy from the [app settings](https://appsignal.com/redirect-to/app?to=edit)

image

Please note that it is an experimental feature so feel free to give that a try and share your feedback

TheFirstAvenger:
@shairyar yes, we have enabled the grouping feature. The issue at hand is the name of the error. While the grouping feature as implemented is correctly grouping based on stack trace location, the name of all of the errors is still a generic name such as `FunctionClauseError`, `CastError`, `ErlangError`, etc... This makes working through the issues list page much harder. This PR modified the name to include the banner, which shows the module/function/line number.

shairyar:
Internal discussion: https://appsignal.slack.com/archives/CNPP953E2/p1646804763995469 (Private link)

unflxw:
Hi @TheFirstAvenger,

We've been discussing this issue today. If we understand correctly, you want to show the full message in the "Issue list" table view, to make it easier to tell the issues apart and find the one you're looking for at a glance. An option we considered was to show, in each table row, a sample error message from one of the errors in the group. Something like this:

Screenshot 2022-03-14 at 13 09 36

(The above is just a quick Web Inspector mockup, so the design is not in any way final -- just illustrating the idea)

Are we understanding the problem right? Would something like the above mockup improve the table for your use case?

shairyar:
@TheFirstAvenger I just wanted to let you know that we will be working on improving it. We will update this ticket once the improvement has been introduced.

PR #665, opened 2021-08-17
Replace decorator slashes with underscores

jeffkreeftmeijer:
Because the Span API uses a Span's name as what was previously described
as the event name, underscores are no longer supported.

Although this was the (documented) case before, changes in the
processing API caused this to break for people using the decorators
(which are a fallback for users upgrading from 1.0).

This patch changes the slashes in automatically generated span names to
underscores, as that's the only available alternative.

It's marked as a minor patch in the changelog, as this changes some
reported names, causing newly created incidents for insidents that
already existed before.

https://docs.appsignal.com/api/event-names.html#event-namin

Closes https://github.com/appsignal/support/issues/132

### DIFF
--- a/.changesets/use-underscores-instead-of-slashes-in-spans-created-from-decorators.md
+++ b/.changesets/use-underscores-instead-of-slashes-in-spans-created-from-decorators.md
@@ -1,0 +1,5 @@
+---
+bump: "minor"
+---
+
+Use underscores instead of slashes in spans created from decorators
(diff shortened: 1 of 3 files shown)


tombruijn (approved):
Please also add a changeset :)

thijsc (approved):

luismiramirez (approved):

jeffkreeftmeijer:
The CI failures aren’t caused by this patch, and are present on main. There’s a fix in #666, please merge that one first.

Issue #685, opened 2021-10-05
Add scheduler utilization to Erlang magic dashboard


jeffkreeftmeijer:
https://blog.appsignal.com/2020/02/04/elixir-monitoring-erlangvm-with-magic-dashboards.html
https://erlang.org/doc/apps/observer/observer_ug.html

- [x] Report scheduler utilization (https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/probes/erlang_probe.ex)
- [ ] Add to magic dashboard (https://github.com/appsignal/public_config/)

unflxw:
Due to the changes to the minutely probes introduced in the appsignal-elixir PR, this documentation page needs to be amended: https://docs.appsignal.com/elixir/instrumentation/minutely-probes.html

PR #708, opened 2021-11-04
Implement scheduler utilization metric

unflxw:
This PR is part of #685. It is a re-do of #687, with an approach based on the existing, tried-and-tested `Appsignal.Probes` implementation. The feedback in appsignal/public_config#33 has been taken into account, and the total utilization metric has been removed.

This commit implements the Erlang scheduler utilization metric on the minutely probes.

In order to report on the scheduler utilization, the probe must keep state from its previous execution. That is, to measure the scheduler utilization between points in time A and B, you must call `:scheduler.sample/0` on point A, then pass the sample it returns to `:scheduler.utilization/1 on point B.

To be able to keep state in the probe, we keep a map of states in `Appsignal.Probes` alongside the map of probes. If a probe takes an argument, the state returned by its previous run is retrieved from the states map and passed as the argument, and the probe's return value is stored in the states map to be used as the next run's state.

### DIFF
--- a/.changesets/add-scheduler-utilization-to-minutely-probes.md
+++ b/.changesets/add-scheduler-utilization-to-minutely-probes.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+---
+
+Add the Erlang scheduler utilization to the metrics reported by the minutely probes. The metric is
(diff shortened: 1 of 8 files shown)


unflxw:
Closing because Lintje does not like the branch name. Reopening as #710. Third time's the charm!

PR #715, opened 2021-11-10
Remove usage of `System.stacktrace/0`

unflxw:
The `System.stacktrace/0` method has been deprecated for a while, and the latest changes in Elixir's `main` branch change it so that it always returns an empty list.

We use `System.stacktrace/0` on the tests for our macro `Appsignal.Stacktrace.get/0` implementation, which itself either calls `:erlang.get_stacktrace/0` or reads the `__STACKTRACE__` variable, to compare its return value against it. The test has instead been modified to check for the presence of some known values in the stacktrace under test.


tombruijn (approved):
Looks fine. We don't need to make sure it matches exactly.

PR #724, opened 2021-11-23
Fix OTP 20 tests

unflxw:
[skip changeset]

This PR must be rebased against `main` after #718 is merged.  The CI checks currently do not pass for the following reason:

- [ ] The integration does not pass the tests under OTP 20.

### Add OTP 20 checks for CI

This commit adds back CI checks for OTP 20 on Elixir 1.9.

### Install Telemetry 0.x version for OTP < 20

The integration compiles successfully when the Telemtry dependency version is set to `~> 0.4`, instead of to `~> 0.4 or ~> 1.0`.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -117,6 +117,10 @@ blocks:
             - ERLANG_VERSION=21.3 ELIXIR_VERSION=1.9.4 . bin/setup
             - mix compile
             - mix test --no-compile
+        - name: Elixir 1.9.4, OTP 20
+          commands:
+            - ERLANG_VERSION=20.3 ELIXIR_VERSION=1.9.4 . bin/setup
+            - mix test
       env_vars:
         - name: MIX_ENV
           value: test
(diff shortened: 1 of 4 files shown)


jeffkreeftmeijer:
I’ve looked at the failure in https://appsignal.semaphoreci.com/jobs/6c7da59f-e934-4f00-888e-3d202625f07f.

It turns out that the pid that’s ignored in the test isn’t the pid that’s actually used. That’s a difference in the OTP versions that doesn’t seem to affect the actual working of the feature, but mismatches with the way the test is set up. I’ve spent some time figuring this out, but haven’t been able to get it to work.

Since Elixir 1.9 is the only version that supports OTP 20, I suggest we discuss disabling this specific test, and dropping OTP 20 in a next minor release.

jeffkreeftmeijer:
Merged #748 in, which fixes the OTP 20 error backend test, so this one’s ready for review.

jeffkreeftmeijer:
@unflxw could you also take another look at this one? I can’t ask you for a review, because this is your pull request. :)

unflxw:
@jeffkreeftmeijer Looks good to me 👍 The branch needs rebasing, though.

jeffkreeftmeijer:
Rebased!

tombruijn (approved):

luismiramirez (approved):

PR #729, opened 2021-11-26
Fix Phoenix parameter filtering

### LINKED ISSUE #649: all request_headers are filtered out by default

unflxw:
Before this change, all sample data was passed through the `Appsignal.Utils.MapFilter.filter/1` function, which filters the given map by the allowlist specified in the `:filter_parameters` Phoenix configuration key, when it exists and is an allowlist.

This commit changes the behaviour of the `set_sample_data/3` function so that it only filters the sample data when the given key is `"params"`. This fixes #649.

### DIFF
--- a/lib/appsignal/span.ex
+++ b/lib/appsignal/span.ex
@@ -190,18 +190,23 @@ defmodule Appsignal.Span do
       |> Appsignal.Span.set_sample_data("environment", %{"method" => "GET"})
 
   """
-  def set_sample_data(%Span{reference: reference} = span, key, value)
-      when is_binary(key) and is_map(value) do
-    data =
-      value
-      |> Appsignal.Utils.MapFilter.filter()
-      |> Appsignal.Utils.DataEncoder.encode()
+  def set_sample_data(span, "params", value) do
+    do_set_sample_data(span, "params", Appsignal.Utils.MapFilter.filter(value))
+  end
+
+  def set_sample_data(span, key, value) do
+    do_set_sample_data(span, key, value)
+  end
+
+  defp do_set_sample_data(%Span{reference: reference} = span, key, value)
+       when is_binary(key) and is_map(value) do
+    data = Appsignal.Utils.DataEncoder.encode(value)
 
     :ok = Nif.set_span_sample_data(reference, key, data)
     span
   end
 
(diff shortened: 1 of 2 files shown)


tombruijn (approved):

lib/appsignal/span.ex
```diff
@@ -190,18 +190,23 @@ defmodule Appsignal.Span do
...
 
   """
-  def set_sample_data(%Span{reference: reference} = span, key, value)
-      when is_binary(key) and is_map(value) do
-    data =
-      value
-      |> Appsignal.Utils.MapFilter.filter()
-      |> Appsignal.Utils.DataEncoder.encode()
+  def set_sample_data(span, "params" = key, value) do
+    do_set_sample_data(span, key, Appsignal.Utils.MapFilter.filter(value))
```
jeffkreeftmeijer, line 194: When matching on `"params"`, we can use `"params"` in the function clause’s body instead of putting it into a variable.

jeffkreeftmeijer (approved):

PR #737, opened 2021-12-01
Fix Telemetry local function warning

unflxw:
When a local function (meaning, either an anonymous function, or a capture of a function without the module being specified) is passed as an event handler to `:telemetry.attach/4` on Telemetry version 1.0.0, it produces a warning about the performance penalty of doing so, due to how these functions are handled by the Erlang VM.

This commit changes the event handler function passed to `:telemetry.attach/4` so that the module is specified in the function capture.

### DIFF
--- a/.changesets/fix-telemetry-local-function-warning.md
+++ b/.changesets/fix-telemetry-local-function-warning.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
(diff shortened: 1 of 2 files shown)


tombruijn (approved):
Looks good! Looks like the same fix as in https://github.com/appsignal/appsignal-elixir-phoenix/pull/35. Any other places we can run into this same issue?

unflxw:
@tombruijn Good thinking! I just grepped for `:telemetry.attach` across the codebase and the one in this PR seems to be the only usage of a local function left.

luismiramirez (approved):

PR #764, opened 2022-02-16
Tracer can ignore a specific process

cblavier:
`Tracer.ignore_process/0` can only ignore current pid.
I want to make Tracer also ignore another pid (see: https://github.com/appsignal/appsignal-elixir/pull/549)


cblavier:
Hi @jeffkreeftmeijer 👋 
I just pushed some tests and the requested refactoring.
Let me know!

PR #776, opened 2022-06-09
Handle empty configurations for all config options

jeffkreeftmeijer:
After the `without_config/0` option was added in #775, I went over the configuration options that use `Application.fetch_env!/2`, added tests to make sure they function without any configuration loaded, and switched them over to `Application.get_env/3`.

### DIFF
--- a/lib/appsignal/config.ex
+++ b/lib/appsignal/config.ex
@@ -162,7 +162,7 @@ defmodule Appsignal.Config do
   end
 
   def request_headers do
-    Application.get_env(:appsignal, :config, %{request_headers: []})[:request_headers] || []
+    Application.get_env(:appsignal, :config, @default_config)[:request_headers] || []
   end
 
   def ca_file_path do
--- a/lib/appsignal/config.ex
+++ b/lib/appsignal/config.ex
@@ -416,7 +416,7 @@ defmodule Appsignal.Config do
         value
 
       :error ->
-        config = Application.fetch_env!(:appsignal, :config)
+        config = Application.get_env(:appsignal, :config, %{})
         value = do_log_file_path(config[:log_path])
         Application.put_env(:appsignal, :"$log_file_path", value)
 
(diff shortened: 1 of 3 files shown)


lib/appsignal/config.ex
```diff
@@ -162,11 +162,12 @@ defmodule Appsignal.Config do
   end
 
   def request_headers do
-    Application.fetch_env!(:appsignal, :config)[:request_headers] || []
+    Application.get_env(:appsignal, :config, %{request_headers: []})[:request_headers] || []
   end
 
   def ca_file_path do
-    Application.fetch_env!(:appsignal, :config)[:ca_file_path]
+    config = Application.get_env(:appsignal, :config, %{ca_file_path: default_ca_file_path()})
```
unflxw, line 169: Would it be possible to add `default_ca_file_path()` as the default value for `:ca_file_path` in `@default_config`, and fetch it from there?

unflxw (approved):

luismiramirez (approved):

lib/appsignal/config.ex
```diff
@@ -162,11 +162,12 @@ defmodule Appsignal.Config do
   end
 
   def request_headers do
-    Application.fetch_env!(:appsignal, :config)[:request_headers] || []
+    Application.get_env(:appsignal, :config, %{request_headers: []})[:request_headers] || []
```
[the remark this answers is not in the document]
jeffkreeftmeijer: You’re right! Fixed. :)

lib/appsignal/config.ex
jeffkreeftmeijer: We’ve been doing it this way since https://github.com/appsignal/appsignal-elixir/commit/13b823174, which introduces “config sources”. Turns out the runtime configuration there has things we can’t hardcode into the compile-time config.

tombruijn (approved):

PR #788, opened 2022-08-01
Always return the span from Span setter functions

jeffkreeftmeijer:
The set_name/2, set_namespace/2 and set_sample_data/3 functions all had
a rescue clause that returned a nil value, even if a span was passed to
them.

This caused issues with optional arguments, where a value that should be
a string was a nil. Instead of returning the span unchanged (which is
expected when chainging functions), these functions would then return a
nil, breaking the setters for the rest of the chain.

This patch makes sure the span is returned as passed, if it's a nil or a
proper span.

### DIFF
--- a/lib/appsignal/span.ex
+++ b/lib/appsignal/span.ex
@@ -104,7 +104,7 @@ defmodule Appsignal.Span do
     end
   end
 
-  def set_name(_span, _name), do: nil
+  def set_name(span, _name), do: span
 
   @spec set_namespace(t() | nil, String.t()) :: t() | nil
   @doc """
@@ -123,7 +123,7 @@ defmodule Appsignal.Span do
     span
   end
 
-  def set_namespace(_span, _name), do: nil
+  def set_namespace(span, _name), do: span
 
   @spec set_attribute(t() | nil, String.t(), String.t() | integer() | boolean() | float()) ::
           t() | nil
@@ -222,7 +222,7 @@ defmodule Appsignal.Span do
     span
   end
 
-  defp do_set_sample_data(_span, _key, _value), do: nil
+  defp do_set_sample_data(span, _key, _value), do: span
 
   @spec add_error(t() | nil, Exception.kind(), any(), Exception.stacktrace()) :: t() | nil
   @doc """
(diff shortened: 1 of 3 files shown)


tombruijn (approved):

test/appsignal/span_test.exs
```diff
@@ -180,16 +180,14 @@ defmodule AppsignalSpanTest do
   end
 
   describe ".set_name/2, with a span that doesn't have a reference" do
-    setup do
-      [return: Span.set_name(%Span{}, "test")]
-    end
+    setup :create_root_span
```
unflxw, line 183: If this is using the output of `:create_root_span` as the original span, is it still testing the case where the span passed is a span that has no reference, as the description of the test case says?

test/appsignal/span_test.exs
```diff
@@ -180,16 +180,14 @@ defmodule AppsignalSpanTest do
...
-    test "returns nil", %{return: return} do
-      assert return == nil
+    setup %{span: span} do
+      [span: span, return: Span.set_name(span, "test")]
     end
 
-    test "does not set the name through the Nif" do
-      assert Test.Nif.get(:set_span_name) == :error
+    test "returns the span", %{span: span, return: return} do
+      assert return == span
```
unflxw, line 190: I don't understand the changes to this test. Do we not want to still test that the name is not set through the Nif?

Does the added setup block do something I'm missing, compared to doing `assert span == Span.set_name(span, "test")`?

lib/appsignal/span.ex
```diff
@@ -222,7 +222,7 @@ defmodule Appsignal.Span do
     span
   end
 
-  defp do_set_sample_data(_span, _key, _value), do: nil
```
unflxw, line 225: Since all branches of `set_sample_data/3` and its related private functions (`set_sample_data/4`, `do_set_sample_data/3`) end up returning `span` in the end in one way or another, instead of returning `span` from many points in the private functions, we could return `span` only once, at the end of `set_sample_data/3`.

unflxw (commented):
✅ on the implementation changes, 🧠❓ on the test changes.

luismiramirez (approved):

jeffkreeftmeijer:
Great finds, @unflxw. Updated!

unflxw (approved):

PR #801, opened 2022-11-11
Logging feature

### LINKED ISSUE #799: Logging
Support for AppSignal logging from the Elixir integration. See https://github.com/appsignal/appsignal-ruby/pull/890 for an exampe implementation.

luismiramirez:
## [Rename Logger to IntegrationLogger](https://github.com/appsignal/appsignal-elixir/commit/e1d72e52529608985fde2ed127b4406433263081)

In order to implement the Logger that comunicates with the Logging API
in the agent, the integration internal logger is renamed to Integration
Logger. This is based in the Ruby and Node.js implementation.

## [Implement Logger module](https://github.com/appsignal/appsignal-elixir/commit/cdf4f5b0da5650a952f0c335dbe8d9f92d04ed70)

This module exposes functions to log messages by severity through the
extension Logging API

## [Add logging endpoint config option](https://github.com/appsignal/appsignal-elixir/pull/801/commits/1815e7fc30aeb9b20296a92c901871b2f2fc4c73)

This option tells the agent where to send the logs made through its API.

## [Update diagnose tests](https://github.com/appsignal/appsignal-elixir/pull/801/commits/6a6a906740a0fd24688a2bcbef0b03e1d509d5ad)

Adds the diagnose tests updated submodule with support for the
`logging_endpoint` config option.

Closes #799

### DIFF
--- a/lib/appsignal/logger.ex
+++ b/lib/appsignal/logger.ex
@@ -1,0 +1,52 @@
+defmodule Appsignal.Logger do
+  require Appsignal.Utils
+
+  @nif Appsignal.Utils.compile_env(:appsignal, :appsignal_tracer_nif, Appsignal.Nif)
+  @severity %{
+    trace: 1,
+    debug: 2,
+    info: 3,
+    warn: 5,
+    error: 6
+  }
+
+  @type log_level :: :trace | :debug | :info | :warn | :error
+
+  @spec trace(String.t(), String.t(), %{}) :: :ok
+  def trace(group, message, metadata \\ %{}) do
+    log(:trace, group, message, metadata)
+  end
+
+  @spec debug(String.t(), String.t(), %{}) :: :ok
+  def debug(group, message, metadata \\ %{}) do
+    log(:debug, group, message, metadata)
+  end
+
+  @spec info(String.t(), String.t(), %{}) :: :ok
+  def info(group, message, metadata \\ %{}) do
+    log(:info, group, message, metadata)
+  end
+
+  @spec warn(String.t(), String.t(), %{}) :: :ok
+  def warn(group, message, metadata \\ %{}) do
+    log(:warn, group, message, metadata)
+  end
+
+  @spec error(String.t(), String.t(), %{}) :: :ok
+  def error(group, message, metadata \\ %{}) do
+    log(:error, group, message, metadata)
+  end
+
+  @spec log(log_level(), String.t(), String.t(), %{}) :: :ok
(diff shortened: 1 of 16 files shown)


thijsc (approved):

tombruijn (approved):

jeffkreeftmeijer (approved):

lib/appsignal/logger.ex
```diff
@@ -1,120 +1,52 @@
...
-          do_log(:stderr, level, message)
-
-        {:file, true} ->
-          do_log(:file, level, message)
-          do_log(:stderr, level, message)
-      end
-    end
+  @spec trace(String.t(), String.t(), %{}) :: :ok
+  def trace(group, message, metadata \\ %{}) do
+    log(:trace, group, message, metadata)
```
unflxw, line 17: For Ruby, we support the log levels that the standard Ruby logger support. (For Node.js, there's no standard logger, so we use the Rust ones instead)

Would it be possible to use for Elixir [the standard log levels that the Elixir logger supports](https://hexdocs.pm/logger/1.14/Logger.html#module-levels)? This would mean removing `trace` and adding a bunch of missing ones. The levels should already [be supported by the agent logging PR](https://github.com/appsignal/appsignal-agent/pull/847/files#diff-31519f6f697f6242b33b48dbf34dc2309d6048713669a0bcf8a8249b9cdb7b85R76-R83), except that `emergency` is called `fatal`.

unflxw (approved):

lib/appsignal/logger.ex
luismiramirez: I assumed we were going for the same levels for the three integrations as they will map to what is seen in AppSignal. I'll remap them to the language standard then.

luismiramirez: Done! Thank you.

unflxw: No prob! Also, I'd be happy to reconsider the log levels for Node.js -- the whole "there's no standard logger so I'll use Rust's levels" is just something I came up with.

Maybe we want to consider `console` the standard logger?

luismiramirez: These are the npm "official" log levels, maybe we can use them.

```js
{
  error: 0,
  warn: 1,
  info: 2,
  http: 3,
  verbose: 4,
  debug: 5,
  silly: 6
}
```

Ref: https://docs.npmjs.com/cli/v8/using-npm/logging

PR #803, opened 2022-11-15
Fix plug dependency for Elixir 1.9

luismiramirez:
Newer plug versions (starting 1.14) require Elixir 1.10 to work. This commit fixes the plug version to the latest 1.13 for older Elixir versions.

[skip changeset]


luismiramirez:
Not sure if we want to release this, as it only affects tests. wdyt?

tombruijn (approved):
No need to release this. Our end-users won't be affected by this change.

PR #804, opened 2022-11-15
Stack trace argument types

jeffkreeftmeijer:
Instead of stripping out arguments from exceptions like `UndefinedFunctionError`s, this patch replaces the arguments with their types where required.

For example, a string (like `”foo”`) is replaced with `binary`. Types like integers aren’t replaced, as they don’t pose a threat to expose sensitive data. For maps and structs, the values are stripped out, but their keys aren’t.

Closes https://github.com/appsignal/support/issues/221.

[skip blogpost]

### DIFF
--- a/test/appsignal/stacktrace_test.exs
+++ b/test/appsignal/stacktrace_test.exs
@@ -33,26 +33,24 @@ defmodule Appsignal.StacktraceTest do
       all_location = expected_location ++ [error_info: %{module: Exception}]
       assert Enum.all?(location, &Enum.member?(all_location, &1))
     end
+
+    test "formats stacktrace lines", %{stack: stack} do
+      [line | _] = stack
+      assert Stacktrace.format([line]) == [Exception.format_stacktrace_entry(line)]
+    end
   end
 
-  test "formats stacktrace lines" do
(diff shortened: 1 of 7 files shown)


tombruijn (approved):

test/appsignal/stacktrace_test.exs
```diff
@@ -33,26 +33,24 @@ defmodule Appsignal.StacktraceTest do
...
+  describe "get/0, with an exception with included arguments" do
+    setup do
+      String.to_atom("string", :extra_argument)
+    catch
+      :error, _ -> %{stack: Stacktrace.get()}
+    end
 
-    [line] = Stacktrace.format(stacktrace)
-    assert line =~ ~r{\(stdlib( [\w.-]+)?\) erl_internal.erl:212: :erl_internal.op_type/2}
+    test "replaces arguments with arities", %{stack: stack} do
```
luismiramirez, line 50:
```suggestion
    test "replaces arguments with types", %{stack: stack} do
```

luismiramirez (approved):
LGTM apart from that tiny nitpick I commented

PR #821, opened 2023-02-02
Reimplement brittle probes tests

jeffkreeftmeijer:
This patch reimplements the probes tests, so they’re no longer dependent on the FakeProbe module, which caused issues in the past. This fixes the brittleness in the ProbesTest, as shown in the [brittle](https://appsignal.semaphoreci.com/branches/f2dc647d-3c7d-45b7-8641-9b4916196050) branch, from which these fixes are extracted.

[skip changeset]


unflxw (approved):
Thank you so much for looking into these tests and fixing them! 💯🎉

luismiramirez (approved):

PR #829, opened 2023-03-02
Logger backend

### LINKED ISSUE #800: Logging backend
Provide a [logging backend](https://hexdocs.pm/logger/Logger.html#module-backends) so users can send their app logs to AppSignal from the configured Logger(s). This will a manual addition to their app config. Not added automatically.

# TODO

- [x] Open pull request (#829)
- [ ] Open docs pull request

jeffkreeftmeijer:
With the new logger backend, users can redirect their logs to AppSignal by attaching the backend in their application.ex files:

``` elixir
Logger.add_backend({Appsignal.Logger.Backend, [group: "phoenix"]})
```

Closes #800.

### DIFF
--- a/lib/appsignal/logger/backend.ex
+++ b/lib/appsignal/logger/backend.ex
@@ -1,20 +1,22 @@
 defmodule Appsignal.Logger.Backend do
   @behaviour :gen_event
 
-  def init({__MODULE__, group}) do
-    {:ok, group}
+  def init({__MODULE__, options}) do
+    {:ok, Keyword.merge([group: "app"], options)}
   end
 
-  def init(_) do
-    {:ok, "app"}
+  def handle_event({level, _gl, {Logger, message, _timestamp, metadata}}, options) do
+    Appsignal.Logger.log(
+      level,
+      options[:group],
+      IO.chardata_to_string(message),
+      Enum.into(metadata, %{})
+    )
+
+    {:ok, options}
   end
 
-  def handle_event({level, _gl, {Logger, message, _timestamp, metadata}}, group) do
-    Appsignal.Logger.log(level, group, IO.chardata_to_string(message), Enum.into(metadata, %{}))
-    {:ok, group}
-  end
-
-  def handle_call(_messsage, group) do
-    {:ok, nil, group}
+  def handle_call(_messsage, options) do
+    {:ok, nil, options}
   end
 end
--- a/lib/appsignal/logger/backend.ex
+++ b/lib/appsignal/logger/backend.ex
@@ -6,7 +6,7 @@ defmodule Appsignal.Logger.Backend do
   end
 
   def init(_) do
-    {:ok, "phoenix"}
+    {:ok, "app"}
   end
 
   def handle_event({level, _gl, {Logger, message, _timestamp, metadata}}, group) do
--- a/lib/appsignal/logger/backend.ex
+++ b/lib/appsignal/logger/backend.ex
@@ -1,0 +1,20 @@
+defmodule Appsignal.Logger.Backend do
+  @behaviour :gen_event
+
+  def init({__MODULE__, group}) do
+    {:ok, group}
+  end
+
+  def init(_) do
+    {:ok, "phoenix"}
+  end
+
+  def handle_event({level, _gl, {Logger, message, _timestamp, metadata}}, group) do
+    Appsignal.Logger.log(level, group, IO.chardata_to_string(message), Enum.into(metadata, %{}))
+    {:ok, group}
+  end
+
+  def handle_call(_messsage, group) do
+    {:ok, nil, group}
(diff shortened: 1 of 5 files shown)


unflxw:
> With the new logger backend, users can redirect their logs to AppSignal by attaching the backend in their application.ex files:
> ```elixir
> Logger.add_backend({Appsignal.Logger.Backend, "phoenix”})
> ```

It should also be possible to configure it in the application environment, like in `config.exs` or `runtime.exs`, right? Quoting [the docs for `add_backend/2`](https://hexdocs.pm/logger/1.12.3/Logger.html#add_backend/2):

```elixir
config :logger, :backends, [{Appsignal.Logger.Backend, "phoenix"}]
```

I'm asking specifically about this because I get the impression that this would be the preferred usage pattern, and I think we should prioritise this in documentation over `add_backend/2` (please let me know if you disagree!)

lib/appsignal/logger/backend.ex
```diff
@@ -0,0 +1,20 @@
...
+  def init({__MODULE__, group}) do
+    {:ok, group}
+  end
+
+  def init(_) do
+    {:ok, "phoenix"}
+  end
+
+  def handle_event({level, _gl, {Logger, message, _timestamp, metadata}}, group) do
+    Appsignal.Logger.log(level, group, IO.chardata_to_string(message), Enum.into(metadata, %{}))
```
unflxw, line 13: Note that, in addition to any user-provided metadata, the message comes with some Elixir-provided metadata: https://hexdocs.pm/logger/1.12.3/Logger.html#module-metadata

It seems to me that some of these will not make it through to AppSignal, because they're composite values (like `crash_reason` or `domain` -- maybe also `mfa`? the docs aren't clear on the format)

It may be interesting to "massage" known ones, or at least those that we consider useful, into strings (probably by calling `inspect()` on them, or maybe something specific to `mfa` so it shows up as `"Module.function/arity"`?)

lib/appsignal/logger/backend.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Appsignal.Logger.Backend do
+  @behaviour :gen_event
+
+  def init({__MODULE__, group}) do
```
unflxw, line 4: In the Node.js instrumentation, we used positional arguments to set the group, and now when adding the format the signature has become a bit unwieldy. Could we use an options list here, to get ahead of that for future extensibility?

(It also makes it clearer what the "phoenix" in `{Appsignal.Logger.Backend, "phoenix"}` means!)

```suggestion
  def init({__MODULE__, [groupname: group]}) do
```

unflxw (commented):
What happens in a scenario where AppSignal is not active? Say, if this backend is configured in the general config, but AppSignal is only active in certain environments. I think the call to `Appsignal.Nif` might break, but I don't really know enough about the Nif setup to be sure.

jeffkreeftmeijer:
> > With the new logger backend, users can redirect their logs to AppSignal by attaching the backend in their application.ex files:
> > ```elixir
> > Logger.add_backend({Appsignal.Logger.Backend, "phoenix”})
> > ```
> 
> It should also be possible to configure it in the application environment, like in `config.exs` or `runtime.exs`, right? Quoting [the docs for `add_backend/2`](https://hexdocs.pm/logger/1.12.3/Logger.html#add_backend/2):
> 
> ```elixir
> config :logger, :backends, [{Appsignal.Logger.Backend, "phoenix"}]
> ```
> 
> I'm asking specifically about this because I get the impression that this would be the preferred usage pattern, and I think we should prioritise this in documentation over `add_backend/2` (please let me know if you disagree!)

Yes, and agreed.

jeffkreeftmeijer:
> What happens in a scenario where AppSignal is not active? Say, if this backend is configured in the general config, but AppSignal is only active in certain environments. I think the call to `Appsignal.Nif` might break, but I don't really know enough about the Nif setup to be sure.

If AppSignal is not active, [the log/2 function is a noop](https://github.com/appsignal/appsignal-elixir/blob/logger-backend/lib/appsignal/nif.ex#L491-L493), so that’s accounted for. :)

lib/appsignal/logger/backend.ex
jeffkreeftmeijer: The metadata gets filtered through https://github.com/appsignal/appsignal-elixir/blob/logger-backend/lib/appsignal/utils/data_encoder.ex, and seems to come in either as types we can handle, or inspected as strings. Although we don’t have specific handling, this should be working like we expect.

unflxw: Awesome! I didn't notice that the metadata went through the data encoder. Marking this as resolved then.

unflxw (approved):

luismiramirez (approved):

PR #830, opened 2023-03-09
Add support for logfmt

jeffkreeftmeijer:
Part of #800 (and based on #829) , this patch adds support for logfmt. 

[skip changeset]


thijsc (approved):

jeffkreeftmeijer:
> Looks good to me.
> 
> I see that this adds support to logfmt for the logging backend, but not for the logger itself -- at least not through the convenience methods (`debug/3`, `info/3` and so on)
> 
> I take it that this is intentional, because there's no known use case for it?

Correct, I think that’s a stretch. That’d require you to explicitly call the log function from your app and pass the logs in the logfmt format. We can always add these later if that’s requested.

tombruijn (approved):

PR #847, opened 2023-05-29
Implement Tesla instrumentation

unflxw:
Part of https://github.com/appsignal/support/issues/257. See [test setup](https://github.com/appsignal/test-setups/pull/113) and [server changes](https://github.com/appsignal/appsignal-server/pull/9562) and [diagnose changes](https://github.com/appsignal/diagnose_tests/pull/87).

Implement an instrumentation for the Tesla HTTP client library, based on our existing Finch instrumentation.

If the Tesla.Middleware.PathParams middleware is used, a template is used as the URL, and parameters are passed separately, to be interpolated by this middleware. If the instrumentation detects that the PathParams middleware is being used correctly, it will send the template URL as part of the sanitised URL. Otherwise, it will only send the scheme, host and port of the URL.

If the Tesla.Middleware.BaseUrl middleware is used, it will be used to form the sanitised URL. Its path will be included in the sanitised URL, as it is assumed to be static. If the PathParams middleware is also being used, the template URL's path fragment will be appended to the base URL. This allows for some "better than nothing" grouping for users who use BaseUrl but not PathParams (e.g. `github.com/users`, instead of just `github.com`)

Caveats
-------

This instrumentation relies on Tesla.Middleware.Telemetry being used before any other middleware. The instrumentation benefits from seeing what the user-intended request was, before the middleware modifies it. Using other middlewares before Telemetry degrades its functionality.

If PathParams is used before Telemetry, the template will already have been interpolated, resulting in a high-cardinality URL. To prevent this, the order of the middlewares is taken into account to determine whether PathParams is being used correctly and whether to include the path fragment of the URL as part of the URL.

If BaseUrl is used before Telemetry, it will have no effect, as the URL will have already been prefixed with the base URL. The base URL will be ignored.

If MethodOverride is used before Telemetry, the method shown will be the actual HTTP method sent over the wire (GET or POST) instead of the semantically intended request method, which is sent in the `X-HTTP-Method-Override` header.

Using Tesla.Middleware.Telemetry before any other middlewares prevents these functionality degradations.

### DIFF
--- a/lib/appsignal/tesla.ex
+++ b/lib/appsignal/tesla.ex
@@ -12,7 +12,7 @@ defmodule Appsignal.Tesla do
     handlers = %{
       [:tesla, :request, :start] => &__MODULE__.tesla_request_start/4,
       [:tesla, :request, :stop] => &__MODULE__.tesla_request_stop/4,
-      [:tesla, :request, :exception] => &__MODULE__.tesla_request_exception/4
+      [:tesla, :request, :exception] => &__MODULE__.tesla_request_stop/4
     }
 
     for {event, fun} <- handlers do
@@ -167,13 +167,5 @@ defmodule Appsignal.Tesla do
 
   def tesla_request_stop(_event, _measurements, _metadata, _config) do
     @tracer.close_span(@tracer.current_span())
-  end
-
-  def tesla_request_exception(_event, _measurements, metadata, _config) do
-    @tracer.current_span()
-    |> @span.add_error(metadata[:kind], metadata[:reason], metadata[:stacktrace])
-    |> @tracer.close_span()
-
-    @tracer.ignore()
   end
 end
--- a/lib/appsignal/tesla.ex
+++ b/lib/appsignal/tesla.ex
@@ -1,0 +1,179 @@
+defmodule Appsignal.Tesla do
+  require Appsignal.Utils
+
+  @tracer Appsignal.Utils.compile_env(:appsignal, :appsignal_tracer, Appsignal.Tracer)
+  @span Appsignal.Utils.compile_env(:appsignal, :appsignal_span, Appsignal.Span)
+
+  @moduledoc false
+
+  require Logger
+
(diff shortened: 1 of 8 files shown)


tombruijn (approved):

lib/appsignal/tesla.ex
```diff
@@ -0,0 +1,188 @@
...
+
+      URI.to_string(%URI{scheme: base_scheme, host: base_host, port: base_port, path: full_path})
+    else
+      full_path = if use_full_path, do: path
+
+      URI.to_string(%URI{scheme: scheme, host: host, port: port, path: full_path})
+    end
+  end
+
+  def env_middlewares(env) do
```
jeffkreeftmeijer, line 113: I don't think this function is used anywhere else. Could you make it private?

lib/appsignal/tesla.ex
```diff
@@ -0,0 +1,188 @@
...
+    # unified value shape.
+
+    middlewares
+    |> Enum.map(fn
+      {:fn, _} -> nil
+      {middleware, _, opts} -> {middleware, opts}
+      {middleware, _} -> {middleware, nil}
+      _ -> nil
+    end)
+    |> Enum.reject(&(&1 == nil))
```
jeffkreeftmeijer, line 150: There's a function to check for nils that you could use here: 

``` elixir
Enum.reject(&is_nil/1)
```

jeffkreeftmeijer (approved):

lib/appsignal/tesla.ex
unflxw: Thank you!

lib/appsignal/tesla.ex
unflxw: It's public so that the unit tests can access it, same as `sanitise_url`. I thought that otherwise, testing the edge cases of this function by sending events through Telemetry got a bit unwieldy.

PR #851, opened 2023-06-07
Add elixir 1.15 to CI

jeffkreeftmeijer:
Add Elixir 1.15.0-rc.1 to CI, and fix a compilation issue introduced in Elixir 1.15 by using rebar instead of mix to install the ssl_verify_fun dependency for now.

[skip changeset]


tombruijn (changes requested):
You probably need to rebase to remove some duplicate commits that are already merged on main.

jeffkreeftmeijer:
This should be good to go, with some more changes and some ignored dialyzer warnings. I've opened https://github.com/appsignal/appsignal-elixir/issues/853 to make sure we get to those later.

tombruijn (approved):

unflxw (approved):
Looks good. Thinking about moving the methods that are delegated to `Logger` to something like `Appsignal.Utils.Logger`, just because Appsignal.Utils is a bit of a potpourri. But it's not important.

luismiramirez (approved):

PR #856, opened 2023-06-15
Reimplement argument cleaning

unflxw:
Fixes appsignal/support#254. Makes progress on appsignal/support#246 (making it clearer, hopefully, by replacing `binary` with `"..."`) but that doesn't fix it (a fix would be, in my opinion, to be able to tell which `handle_event` implementation is being called)

---

Hopefully striking a better balance between:
- Showing enough information. It should be possible to often know which function clause is being triggered by a cleaned backtrace line.
- Not showing too much information. Describing every item in a list or a struct results in the backtrace line being truncated, often before getting to the interesting part.
- Not leaking personally identifiable information.

There is no clear "good solution" here. This implementation does the following:
- Avoid expanding lists (unless they're keyword lists)
- Show atoms, booleans and `nil`, as these are understood to have meaning in the codebase itself.
- Do not show strings or numbers or other potentially personally identifiable information.
- Remove details (PIDs, ports, references) that can interfere with backtrace line grouping.
- Describe maps and tuples, but only if they're small and not too nested.
- Do not describe structs, as their structure is already known.
- Try to use syntax that will be familiar to the user when describing backtrace entries (similar to `inspect()` if possible, similar to typespecs otherwise)


unflxw:
Fixes appsignal/support#254. Makes progress on appsignal/support#246 (making it clearer by replacing `binary` with `"..."`) but that doesn't fix it (a fix would be, in my opinion, to be able to tell which `handle_event` implementation is being called)

lib/appsignal/utils/argument_cleaner.ex
```diff
@@ -1,17 +1,124 @@
...
+    else
+      "%{#{contents}}"
+    end
+  end
+
+  def clean(argument, _) when is_integer(argument), do: "integer()"
+
+  def clean(argument, _) when is_float(argument), do: "float()"
+
+  def clean(argument, _) when is_pid(argument), do: "#PID<...>"
```
tombruijn, line 37: I don't see this show up. I get `"..."` for pids.

```elixir
  def backtrace_error(_conn, _params) do
    port = Port.open({:spawn, "echo hello"}, [:binary])
    backtrace_error_inner(%{pid: System.pid(), nil: nil, true: true, false: false, int: 123, float: 12.3, abc: "def", ghi: %{jkl: %{mno: "pqr", int: 123, float: 12.3}}}, System.pid(), port, nil, true, false, 123, 12.3, "abc")
  end

  def backtrace_error_inner(%{abc: "def", ghi: %{jkl: %{other: v}}} = arg) do
  end

  def backtrace_error_inner(%{abc: "def", ghi: %{jkl: %{other: v}}} = arg, _pid, _port, _nil, _true, _false, _int, _float, _string) do
  end
```

image

unflxw: The PID data type in Elixir represents the IDs of BEAM "processes" (which are not OS processes, but internal to the VM) -- you would use `self()` to obtain the PID of the current process.

By contrast, `System.pid()` returns the process ID of the OS process that runs the whole instance of the VM. And for some reason, it returns it as a string -- hence why it's omitted as `"..."`.

lib/appsignal/utils/argument_cleaner.ex
```diff
@@ -1,17 +1,124 @@
...
+      Enum.join(Enum.reject(["fn", placeholders, "-> ... end"], &(&1 == "")), " ")
+    else
+      # As it is not an anonymous function, calling `inspect` on it already
+      # expresses it in the Module.function/arity form, which is fine.
+      inspected_function
+    end
+  end
+
+  def clean(argument, recurse) when is_tuple(argument) do
+    if Enum.empty?(Tuple.to_list(argument)) do
```
jeffkreeftmeijer, line 60: There are a couple of functions which have an `if`-statement as their function bodies that have a specific case when the value is empty, which I think could be rewritten to use multiple function clauses.

In this example, because we know what an empty tuple looks like, we can add a clause that returns `"{}"` when the argument is `{}` and circumvent the "main" clause. That would make the code more readable, and save a call to `Tuple.to_list/1`.

``` elixir
  def clean({}, _), do: "{}"

  def clean(argument, recurse) when is_tuple(argument) do
    "{#{do_clean_tuple(argument, recurse)}}"
  end
```

jeffkreeftmeijer (approved):
Loving this patch. Amazing work on the tests, and I think implementing `Inspect` for a custom type is indeed the way to go here. 🙌 

I left some comments that are mostly about the structure of the code. We could use some Elixir-isms to make the code a bit more readable. No blockers there, though.

unflxw:
Thanks @jeffkreeftmeijer! Implementing `Inspect` for the cleaned arguments is itself lifted from your previous implementation (the `Type` struct did that)

I'll take a look at the changes you proposed and try to make some of those functions more lightweight.

lib/appsignal/utils/argument_cleaner.ex
unflxw: Oh, that makes a lot of sense -- I didn't realise that I could check for the zero-member case directly with pattern-matching. Will do!

lib/appsignal/utils/argument_cleaner.ex
```diff
@@ -1,17 +1,124 @@
...
+  def clean(argument, _) when is_reference(argument), do: "#Reference<...>"
+
+  def clean(argument, _) when is_function(argument) do
+    inspected_function = inspect(argument)
+
+    if String.starts_with?(inspected_function, "#Function<") do
+      # Represent the function as `fn _, _ -> ... end`, as this allows
+      # to express the arity clearly while redacting the rest.
+      arity = :erlang.fun_info(argument)[:arity]
+      placeholders = if arity == 0, do: "", else: Enum.map_join(1..arity, ", ", fn _ -> "_" end)
```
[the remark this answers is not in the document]
unflxw: Annoyingly, I have to check for `Enum.empty?` here because, with `1..arity`, `Enum.to_list(1..0)` results in `[1, 0]`, which is not what we want. We could use `1..arity//1` to fix that (`Enum.to_list(1..0//1) == []`) but it's not compatible with Elixir 1.11, which we still support.

I'm using `map_join/3` because it seems like the right tool for the job (and because the previous implementation used it) but I'm not a fan either -- I can never remember the order of the arguments myself.

I'll switch to the second implementation you suggest, the one that uses `map_join/3` with the separate zero-arity clause.

jeffkreeftmeijer:
> Implementing `Inspect` for the cleaned arguments is itself lifted from your previous implementation (the `Type` struct did that)

Oh wow. It deed feel familiar indeed. 🤦 

Also, let me know if I can help updating those functions.

tombruijn (approved):

unflxw:
To do before merging: on OTP 26 ([issue about adding it to CI](https://github.com/appsignal/appsignal-elixir/issues/857)), the ordering of small maps is no longer insertion/declaration order ([blog post about this change](https://fly.io/phoenix-files/taking-control-of-map-sort-order-in-elixir/), sort of). We should sort keys manually. (The tests for small maps should fail on OTP 26 due to this)

lib/appsignal/utils/argument_cleaner.ex
unflxw: Done for tuples and lists. Can't do this for maps, as pattern-matching for maps is non-exhaustive (`%{}` matches any map, not the empty map)

lib/appsignal/utils/argument_cleaner.ex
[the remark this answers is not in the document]
unflxw: Done.

luismiramirez (approved):

PR #870, opened 2023-07-14
Allow JSON and Logfmt log messages

unflxw:
Instead of passing metadata to the convenience methods, a format can be passed which will be used to parse metadata out of the message in the Logfmt or JSON format.

Fixes appsignal/support#244.


luismiramirez (approved):

tombruijn (approved):
Please add a changeset

PR #876, opened 2023-08-08
Add string cast to agent diagnose

luismiramirez:
Explicitly casting the NIF's diagnose return type to string makes dialyzer happy in 1.15.

---

This is clearly a hack-ish workaround. Couldn't get a better solution for it. Another option would be to make the `_diagnose` function from the unloaded NIF return other than an atom, but I think this PR's implementation is cleaner.

Fixes: #853 

[skip changeset]


jeffkreeftmeijer (approved):
It's a creative fix, but since the `to_string` function is a noop (https://github.com/elixir-lang/elixir/blob/v1.12.3/lib/elixir/lib/string/chars.ex#L35-L38), I prefer this over updating the ignored warnings every time the line numbers change. I'm +1.

PR #894, opened 2023-09-26
improvement: add `register_parent/1` to `Appsignal.Tracer`

zachdaniel:
Need something like this for `ash_appsignal` which starts processes for you.

See #895 for context.

### DIFF
--- a/lib/appsignal/tracer.ex
+++ b/lib/appsignal/tracer.ex
@@ -207,6 +207,25 @@ defmodule Appsignal.Tracer do
     :ok
   end
 
+  @doc """
+  Registers a span as the parent span for any spans started by this process.
+
+  This is necessary when you want to instrument asynchronous work.
+
+      parent = Appsignal.Tracer.current_span()
+
+      list
+      |> Task.async_stream(fn item ->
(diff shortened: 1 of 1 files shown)


zachdaniel:
Not sure if I necessarily need to follow those instructions up there ^. Seems weird for me to explicitly tag reviewers, for example, as an outside contributor. Some guidance would be appreciated :)

unflxw:
@zachdaniel Ah, no, don't worry about that. Adding us as reviewers is good, so we don't forget about it. Any other changes that need to be done, like the changeset, we'll likely do them ourselves.

(The bot is also, it seems, confused about whose PR this is, and keeps editing the message mentioning different people 😅)

Re: naming (everyone's favorite topic) I'd go with `register_current/1`, as the span that's being registered is the current span. But anything is fine, I think it's likely we'll `@doc false` this for the time being.

jeffkreeftmeijer (approved):
Looks good! Agreed with @unflxw, could you update the name for the added function and `@doc false` it?

unflxw:
Continuing work at #898; closing.

PR #904, opened 2023-11-21
Always use Jason to encode/decode JSON

unflxw:
Instead of using `Appsignal.Json` to conditionally pick between Poison and Jason, adding the complexity of having to install one of the two alongside AppSignal itself, add Jason as a direct dependency.


tombruijn (approved):
Please add a changeset. Probably a minor version bump.

luismiramirez (approved):

PR #915, opened 2024-01-22
Fix Ecto parallel preloads

### LINKED ISSUE #603: Support for parallel Ecto preloads
Ecto can preload data using separate tasks. Currently we don't connect those spans to the span above, so this detail is lost. It would be good if these could be tracked.

As reported [here](https://elixirforum.com/t/help-understanding-ecto-telemetry-events-according-to-appsignal/35957) on the Elixir forum.

unflxw:
Fixes #603. See also https://github.com/appsignal/appsignal-docs/pull/1715. Use https://github.com/appsignal/test-setups/pull/184 to test it.

Add the `Appsignal.Ecto.Repo` module to fix the Ecto parallel preload issue, making it easy to apply the workaround that was already found by [Elixir Forums][solution] users.

After this change, customers can replace `use Ecto.Repo` in their Ecto repositories with `use Appsignal.Ecto.Repo`. This will configure Ecto so that it passes through the current span at invocation time as the telemetry options for the query, fixing the issue where parallel preload queries were lost because they took place in parallel processes where there was no current span.

For customers who wish to customise their Ecto repo default options, which would override the `default_options/1` function implemented by `Appsignal.Ecto.Repo`, a workaround is provided by also exposing this function as `Appsignal.Ecto.Repo.default_options`, so customers can call it manually and merge its output with their own default options.

[solution]: https://elixirforum.com/t/help-understanding-ecto-telemetry-events-according-to-appsignal/35957/3

### DIFF
--- a/.changesets/add-appsignal-ecto-repo-to-fix-parallel-preloads.md
+++ b/.changesets/add-appsignal-ecto-repo-to-fix-parallel-preloads.md
@@ -3,7 +3,7 @@ bump: "patch"
 type: "add"
 ---
 
-Add `Appsignal.Ecto.Repo` to fix parallel preloads.
+Add `Appsignal.Ecto.Repo` to support parallel preloads.
 
 For AppSignal to be able to instrument parallel preloads, the current instrumentation context needs to be passed from the Elixir process that spawns the preload to the short-lived processes that run each of the parallel queries.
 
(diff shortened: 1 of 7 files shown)


unflxw:
I was wondering if there would be an easy way for `Appsignal.Ecto.Repo` to automatically register the telemetry handlers for the Ecto repo it is used in, providing a slightly easier way to register additional repos than `Appsignal.Ecto.attach/2` and lessening the reliance on the OTP app auto-detection.

(This is not a blocker, just a thought)

luismiramirez (approved):

jeffkreeftmeijer (approved):
Such a nice patch! I'm very excited that this is an add-on instead of a completely new implementation. 👍

tombruijn (approved):

PR #918, opened 2024-02-07
Replace deprecated `use Mix.Config` with `import Config`

nallwhy:


tombruijn (approved):
Should be good to merge, as we only support 1.9 and newer.

> The [Config](https://hexdocs.pm/elixir/main/Config.html#content) module in Elixir was introduced in v1.9 as a replacement to use Mix.Config, which was specific to Mix and has been deprecated.

https://hexdocs.pm/elixir/main/Config.html

luismiramirez (approved):

unflxw (approved):
Thank you for the contribution @nallwhy!

PR #930, opened 2024-03-21
Drop non-500 errors in add_error/3

jeffkreeftmeijer:
Plug sets an attribute named plug_status on each of its raised exceptions, which appsignal-plug uses to filter non-500 errors out. This patch moves that filtering to the add_error/3 function, allowing the integration to properly filter exceptions originating in plug, even if they are not caught by the Plug integration.

### DIFF
--- a/.changesets/drop-non-500-plug-errors-in-add_error-3.md
+++ b/.changesets/drop-non-500-plug-errors-in-add_error-3.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Drop non-500 Plug errors in add_error/3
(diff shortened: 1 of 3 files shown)


luismiramirez (approved):

unflxw (approved):
Approving either way, but I'd like to understand this a bit better. I see [we do the same check in Plug](https://github.com/appsignal/appsignal-elixir-plug/blob/main/lib/appsignal_plug.ex#L128-L130). How come it needs to be done here as well?

jeffkreeftmeijer:
We have a customer that's running into an issue where they're getting a 400-type error reported through the error backend. That doesn't go through the Plug integration, so it's not filtered out. This patch moves it here, and allows us to remove it from the Plug integration. Eventually, we might be able to remove the separate Plug (and Phoenix) integrations altogether, and go back to having one package that doesn't actually depend on either.

More context in this thread on Slack: https://appsignal.slack.com/archives/C7XHXQ7N0/p1710949166329769

PR #947, opened 2024-06-17
Lock credo to 1.7.6 on Elixir < 1.13

jeffkreeftmeijer:
Credo 1.7.7 dropped support for Elixir versions below 1.13. This patch locks the credo version to 1.7.6 on any Elixir version before 1.13.

[skip changeset]

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -132,6 +132,12 @@ defmodule Appsignal.Mixfile do
         _ -> "~> 1.14"
       end
 
+    credo_version =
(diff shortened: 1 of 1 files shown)


tombruijn (approved):

luismiramirez (approved):

unflxw:
✅ Wondering whether it makes sense to only run Credo on the latest stable Elixir? It could be confusing if different versions of Credo give different warnings.

PR #953, opened 2024-06-26
Drop Elixir versions before 1.11

jeffkreeftmeijer:
AppSignal  for Elixir supports the Elixir and OTP versions listed on the [compatibility page](https://hexdocs.pm/elixir/main/compatibility-and-deprecations.html). Currently, that's Elixir 1.13 and up, and therefore OTP 25 and up.

However, we don't tend to remove old versions from CI until they start giving us issues. When switching the CI to GitHub actions, we found that we could no longer install OTP versions lower than 24, as they're incompatible with the Ubuntu version we're running. Instead of trying to compile from source, we decided to drop Elixir 1.9.x and 1.10.x. Currently, the CI runs Elixir 1.11 and up, and OTP 24 and up.

By dropping these versions, we were able to remove some code we added for compatibility with versions before Elixir 1.11:

1. `Appsignal.Utils.warning/1` was added to bridge the gap between `Logger.warn/1` and `Logger.warning/1`, after the rename in 1.11. Since we don't support versions before 1.11 anymore, we don't need the helper function and we can just call `Logger.warning/1` directly.

2. `Appsignal.Utils.compile_env/3` is a similar case, in which `Application.compile_env/3` was introduced in Elixir 1.10. The helper dropped down to  `Application.get_env/3` on older versions. We can now call `Application.compile_env/3` directly.

Both of these functions were also used in AppSignal for Phoenix and AppSignal for Plug. Pull requests are opened for both (https://github.com/appsignal/appsignal-elixir-phoenix/pull/94 and https://github.com/appsignal/appsignal-elixir-plug/pull/47). 

Because this pull request includes fallbacks, we don't need to release these in a specific order. In unlikely cases where a user updates AppSignal for Elixir without updating Appsignal for Phoenix, the integration will remain working, but it will throw deprecation warnings.

### DIFF
--- a/lib/appsignal.ex
+++ b/lib/appsignal.ex
@@ -15,7 +15,6 @@ defmodule Appsignal do
   use Application
   alias Appsignal.Config
   require Logger
-  import Appsignal.Utils, only: [info: 1, warning: 1]
 
   @doc false
   def start(_type, _args) do
@@ -87,7 +86,7 @@ defmodule Appsignal do
   def initialize do
     case {Config.initialize(), Config.configured_as_active?()} do
       {_, false} ->
-        info("AppSignal disabled.")
+        Logger.info("AppSignal disabled.")
 
       {:ok, true} ->
         Appsignal.IntegrationLogger.debug("AppSignal starting.")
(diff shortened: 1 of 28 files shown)


unflxw (approved):

jeffkreeftmeijer:
I've decided to remove the deprecation warnings for now (https://github.com/appsignal/appsignal-elixir/pull/953/commits/3457f08519f912ced749434d1b03461e0b3d2231), to allow us to merge this patch without having to immediately having to release it. We'll add a deprecation warning later, after a new version of the AppSignal for Phoenix and Plug libraries have been released.

luismiramirez (approved):

PR #960, opened 2024-08-29
Check-in scheduler

unflxw:
Part of https://github.com/appsignal/integration-guide/issues/169.

### [Implement NDJSON support in the transmitter](https://github.com/appsignal/appsignal-elixir/commit/ad3220987cbc62adbd8bc19ad5d8f30b7b6314bd)

Modify the `transmit/3` method signature so that, if a payload is
provided, a format for that payload to be encoded with must be
provided alongside it.

Implement support for NDJSON as said format.

### [Fix test warnings](https://github.com/appsignal/appsignal-elixir/commit/abf869662cd60a85706b5e9c20884223703297bb)

Fix a warning on an unused alias and a missing assertion on the tests
for `Appsignal.Ecto.Repo`.

Fix a warning on an unused test helper on the `Appsignal.Absinthe`
tests.

### [Implement check-in scheduler](https://github.com/appsignal/appsignal-elixir/commit/003e97c88793e32447c0c13ebcd0d0043356f58c)

Implement a check-in event scheduler, which schedules check-in events
to be transmitted in a separate Elixir process, with a minimum wait
period of ten seconds between requests, and a wait period of a tenth
of a second before the first request, referred to as "debounce"
periods.

This is a relatively minor improvement for the existing cron
check-ins, but it is a requirement for the heartbeat check-ins, both
to avoid slowing down customers' applications with blocking requests,
and to avoid misuse of the feature from spamming our servers.

The scheduler also acts as a deduplicator, removing "similar enough"
check-in events -- again, not particularly interesting for cron
check-ins, but a requirement to minimise damage when heartbeat
check-ins are misused.

Use the previously implemented support for NDJSON payloads in the
transmitter in order to send all the check-ins in a single request.

Similar to the probes, an Elixir process is started by the AppSignal
application supervisor. When the process receives an event to be
transmitted, it stores it and schedules a message, which will trigger
the transmit, to be sent to itself at a later time.

When this Elixir process receives a shutdown signal, it attempts to
transmit all scheduled events before it shuts down.


tombruijn (approved):
My knowledge of the GenServer specifics is a bit fuzzy at the moment, but looks sane 👍

unflxw:
@tombruijn Thanks! Ping me if you want me to walk you through it. (same goes for anyone else)

PR #967, opened 2024-10-10
Add `reported_by` tag to error backend errors

unflxw:
Motivated by [this Intercom conversation](https://app.intercom.com/a/inbox/yzor8gyw/inbox/admin/5246522/conversation/16410700366093).

We get support tickets about incomplete error reporting, which we assume is due to the use of the error backend. Often customers do not know whether the error was raised from an uninstrumented context, or whether they're using a version of the integration where the error backend is enabled by default.

This should simplify the support workflow for it going forward, by giving us that information in the error sample itself. We can also ask customers to update to the latest version and see if the affected samples have this tag.

This follows the format that we use in [the Ruby integration][ruby].

[ruby]: https://github.com/appsignal/appsignal-ruby/blob/6201f5dbe662e22e682a8a324b7e996e979ef09f/lib/appsignal/integrations/puma.rb#L21

### DIFF
--- a/.changesets/add--reported_by--tag-to-errors-reported-by-the-legacy-error-backend.md
+++ b/.changesets/add--reported_by--tag-to-errors-reported-by-the-legacy-error-backend.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: add
+---
+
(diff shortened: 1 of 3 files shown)


luismiramirez (approved):

unflxw:
This needs a changeset before merging.

PR #982, opened 2025-02-24
Update bundled certificates

luismiramirez:
**Update bundled certificates**
Updated to revision 2024-12-31 of the Mozilla CA certificate store, as published by curl.se: https://curl.se/ca/cacert-2024-12-31.pem

**Limit Hackney version to 1.21.0 max**
Due to an error on Hackney 1.22.0 when sending TLS requests, limit it to
the latest compatible version.


.changesets/limit-hackney-to-1-21-0.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+Hackney compatibility is limited up to 1.21.0.
```
unflxw, line 6: I believe it's more useful for users if the changelog describes the issue that is being fixed.

```suggestion
Fix an issue where the AppSignal agent fails to download during installation, by restricting the Hackney dependency to version 1.21.0 or lower.
```

unflxw (approved):

PR #983, opened 2025-02-26
Switch Hackney to Finch

### LINKED ISSUE #978: Consider allowing finch as an alternative to hackney
The only dependency in our one codebase depending on hackney is appsignal. Hackney currently has an open security issue: https://github.com/advisories/GHSA-vq52-99r9-h5pw

Would it be possible to allow Finch and make hackney an optional dependency as with the Swoosh and AWS packages?

### LINKED ISSUE #981: mix appsignal.install fails due SSL handshake error
## Describe the bug

When trying to run `mix appsignal.install` on a new project, it fails with the following error:

```bash
Validating Push API key:
09:04:06.354 [notice] TLS :client: In state :certify at ssl_handshake.erl:2186 generated CLIENT ALERT: Fatal - Handshake Failure
 - {:bad_cert,
 {:hostname_check_failed, {:requested, {185, 191, 3, 125}},
  {:received, [dNSName: ~c"push.appsignal.com"]}}}

Validating failed, reason:

  {:tls_alert, {:handshake_failure, ~c"TLS client: In state certify at ssl_handshake.erl:2186 generated CLIENT ALERT: Fatal - Handshake Failure\n {bad_cert,\n     {hostname_check_failed,\n         {requested,{185,191,3,125}},\n         {received,[{dNSName,\"push.appsignal.com\"}]}}}"}}

```

## To Reproduce

- Generate a new project using `mix new xyz`
- Add dependency `{:appsignal, "2.13.4"}`
- Run `mix deps.get`
- Run `mix appsignal.install <your-api-key>`

luismiramirez:
**Transition the HTTP client usage from Hackney to Finch to optimize
requests performance.**

Two versions of transmissions are required now, as Finch needs to be run
under the supervision tree. One version is a standalone transmission
that starts and kills the process as soon as the requests is made, and
the other one uses the pool that's created when the integration starts.

Fixes #978
Fixes #981

### DIFF
--- a/.changesets/switch-hackney-with-finch.md
+++ b/.changesets/switch-hackney-with-finch.md
@@ -1,0 +1,6 @@
+---
+bump: minor
+type: change
+---
+
+Switch Hackney to Finch as the bundled HTTP client to operate within the integration
--- a/.changesets/switch-the-internal-http-client--hackney--with-finch.md
+++ b/.changesets/switch-the-internal-http-client--hackney--with-finch.md
@@ -1,0 +1,6 @@
+---
+bump: minor
+type: change
+---
+
+Switch the internal HTTP client (hackney) with Finch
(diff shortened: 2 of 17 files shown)


jeffkreeftmeijer (approved):
Pushed a fix for the AppsignalFinch filter. Aside from that this looks good!

PR #995, opened 2025-04-03
Use Access.get/3 for boolean configuration helpers

jeffkreeftmeijer:
Previously, functions like minutely_probes_enabled?/0 used Map.get/3 to get values from the configuration. In some situations, the configuration is a keyword list. By using Access.get/3 both versions are supported.

### DIFF
--- a/.changesets/allow-keyword-list-configurations.md
+++ b/.changesets/allow-keyword-list-configurations.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: add
+---
+
+Allow keyword list configurations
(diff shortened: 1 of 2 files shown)


unflxw (approved):
Looks good to me. Is there any context to this change that I'm missing? Does it relate to a support case?

luismiramirez (approved):

jeffkreeftmeijer:
I'm still figuring out why this happens exactly, but we have more customers running into this (https://app.intercom.com/a/inbox/yzor8gyw/inbox/admin/540654/conversation/16410700415198). This at least fixes the issue of part of the integration failing to start.

PR #997, opened 2025-04-08
Set prefix attribute in Oban start handler

tfwright:
Just throwing this out there to see if something like this makes sense to add.

I have a multi tenant application that uses an independent Oban instance for each tenant, identified by the `prefix` value in the config. 

In order to include this information in appsignal data, I have needed to manually copy the value into the metadata. By setting it here, it can be automatically included for all jobs.

### DIFF
--- a/lib/appsignal/oban.ex
+++ b/lib/appsignal/oban.ex
@@ -70,6 +70,14 @@ defmodule Appsignal.Oban do
     |> @span.set_attribute("worker", to_string(worker))
     |> @span.set_attribute("appsignal:category", "job.oban")
 
+    # The `:conf` metadata key was added in Oban v2.4.0
+    conf = metadata[:conf]
+
+    if conf && conf.prefix do
+      @span.set_attribute(span, "prefix", to_string(conf.prefix))
+    end
+
+
     # The `:job` metadata key was added in Oban v2.3.1.
     job = metadata[:job]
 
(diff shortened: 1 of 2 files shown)


unflxw:
Hi @tfwright, thank you for your contribution! This seems like a very reasonable suggestion.

The changes in this PR need to be amended so that this specific scenario (the presence of the `conf` key) is tested in the Oban unit tests, as well as its absence.

They also need to account for the possibility that `conf.prefix` isn't present, because no prefix has been set. Depending on the version in which `conf` was added to the metadata, it might need to account for `conf` itself not being present in the metadata as well -- see how this is scenario is handled for the `:job` key: https://github.com/appsignal/appsignal-elixir/blob/807d2b1e2aaab2680e187288fd07649feaebf652/lib/appsignal/oban.ex#L73-L82

tfwright:
@unflxw thanks for the feedback.

I made the addition conditional and added a test trying to match the existing style of the version specific feature tests.

tfwright:
Oof, I was hoping to get away with configuring a local env to test this 😬 

I guess I'll give it one more shot

unflxw:
Don't worry @tfwright! I'm happy to ship it from here if you'd rather 😅

unflxw:
Apologies for the delay @tfwright! Continuing work on #1001. Closing.

PR #1000, opened 2025-04-25
Remove redundant cron check-in pairs

unflxw:
See https://github.com/appsignal/appsignal-ruby/pull/1407 for context. This commit reimplements that in Elixir.

Before submitting a batch of cron check-in events, check for any pairs of events that are redundant with other pairs of cron check-in events. This would happen, for example, when running a `cron` block for a given identifier many times in quick succession.

A redundant pair of cron check-in events is a pair of start and finish cron check-in events for a given identifier, given that another pair of start and finish check-in events exists for that identifier.

### DIFF
--- a/.changesets/remove-redundant-cron-check-in-pairs.md
+++ b/.changesets/remove-redundant-cron-check-in-pairs.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: change
+---
+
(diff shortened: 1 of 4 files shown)


luismiramirez (approved):

tombruijn (approved):
Don't forget to add a changeset :)

unflxw:
@tombruijn Ah, thanks! I'll add one for the Ruby PR as well.

PR #1034, opened 2026-03-12
Add documentation for architectural decisions

jeffkreeftmeijer:
This PR adds documentation for architectural decisions and implementation details that were never formally written down. The documentation focuses on the "why" behind design choices, particularly around the ETS-based tracer registry, NIF integration, configuration system, instrumentation patterns, and error handling.

[skip changeset]


docs/configuration-system.md
```diff
@@ -0,0 +1,467 @@
...
+
+Examples:
+- `active: false` - AppSignal is inactive by default
+- `instrument_ecto: true` - Ecto instrumentation is enabled by default
+- `log: "file"` - Logs go to file by default
+- `send_params: true` - Request parameters are sent by default
+
+### 2. System Detection
+
+Automatically detected system configuration (lib/appsignal/config.ex:268-285):
```
tombruijn: These line number may go out of sync quite quickly. Would it be better to refer to the function names, e.g. `load_from_system`? Here and in other places in these documents.

docs/configuration-system.md
```diff
@@ -0,0 +1,467 @@
...
+  valid?(config)  # Reuses the config
+end
+```
+
+**Why?**
+
+`Application.get_env/2` requires an ETS lookup.
+When checking if AppSignal is active, doing two lookups is wasteful.
+Passing the config forward saves the second lookup.
+
```
tombruijn: Do we need to document small optimizations like these?

docs/error-handling.md
```diff
@@ -0,0 +1,353 @@
...
+Check if enabled (lib/appsignal/config.ex:188-193):
+
+```elixir
+def error_backend_enabled? do
+  case Application.fetch_env(:appsignal, :config) do
+    {:ok, value} -> !!Access.get(value, :enable_error_backend, false)
+    _ -> false
+  end
+end
+```
```
tombruijn: I personally don't need this level of detail. I can find the function in the code.

docs/instrumentation-decorators.md
```diff
@@ -0,0 +1,473 @@
...
+The AppSignal Elixir integration provides two ways to instrument code: functional wrappers and compile-time decorators.
+This document explains both approaches, their design, and important implementation details.
+
+## Overview
+
+Instrumentation creates spans to track execution time and capture contextual information about code execution.
+The integration provides:
+
+1. **Functional instrumentation** - `Appsignal.instrument/2` and friends
+2. **Decorator-based instrumentation** - `@decorate` macros for compile-time wrapping
```
tombruijn, line 12: What's your opinion on the decorators? Are they something we should keep long term of phase out?

docs/nif-rust-integration.md
```diff
@@ -0,0 +1,340 @@
...
+
+  case :erlang.load_nif(path, 1) do
+    :ok ->
+      :ok
+
+    {:error, {:load_failed, reason}} ->
+      arch = :erlang.system_info(:system_architecture)
+
+      IO.warn(
+        "Error loading NIF (Is your operating system (#{arch}) supported? ...
```
tombruijn:
```suggestion
        "Error loading NIF (Is your operating system (#{arch}) supported? ..."
```

docs/nif-rust-integration.md
```diff
@@ -0,0 +1,340 @@
...
+
+- `start()` - Start the agent
+- `stop()` - Stop the agent
+- `diagnose()` - Run diagnostics and return a report
+- `running_in_container?()` - Check if running in a container
+- `loaded?()` - Check if the NIF was successfully loaded
+
+### Logging
+
+- `log(group, severity, format, message, attributes)` - Send a log entry to AppSignal
```
tombruijn: I don't really need this list here. If we need docs for these functions, let's add code comments for their purpose.

docs/span-lifecycle.md
```diff
@@ -0,0 +1,182 @@
...
+
+This uses `Appsignal.Utils.DataEncoder` to convert Elixir terms to the NIF data format.
+
+## Span Closing
+
+Spans must be explicitly closed:
+
+```elixir
+Span.close(span)
+Span.close(span, end_time)
```
tombruijn: It would be nice to have an example or explanation of why sometimes the span time is given.

docs/tracer-registry.md
```diff
@@ -0,0 +1,271 @@
+# Tracer Registry (ETS Table)
+
+The Appsignal Elixir integration uses an ETS table as a registry to track spans across processes.
```
tombruijn:
```suggestion
The AppSignal Elixir integration uses an ETS table as a registry to track spans across processes.
```

docs/tracer-registry.md
```diff
@@ -0,0 +1,271 @@
...
+  schedule_sync()
+  pids = MapSet.new(monitored_pids())
+  {:noreply, pids}
+end
+```
+
+This happens every 60 seconds by default (`@sync_interval`).
+
+**Why?**
+This synchronization ensures the Monitor's internal state stays consistent with the actual Erlang process monitors, recovering from any potential state drift.
```
tombruijn, line 94: This is for clean up of no longer instrumented processes?

jeffkreeftmeijer:
Alright, this was a bit of a tryout to see if we can generate documentation based on the code and commit history of the project (as requested). I believed we could get pretty good results, as most of the decisions have been documented in the commit messages, so I instructed Claude to go over my commits to gain context and write documentation.

However, what was generated was mostly a description of the code as is. The result is so tightly coupled (including code samples and direct line locations in the code), that it'd surely become a maintenance burden if it was merged as is. Aside from that, it's very verbose in explaining parts of the code that aren't interesting, and it lacks or ignores context that would be. Instead of documenting why decisions were made, it mostly explains what the code does, which one can also find when reading the code.

I wasn't able to get through it yet (it's quite a bit of work), but I'm working on a patch on top of this to remove the cruft and add some valuable context.

This pull request is interesting as a case study. AI-generated documentation often appears valuable, but because it’s produced directly from the code, it typically ends up as little more than another representation of that same code. Also, the verbosity of the result dissuades anyone from actually reading it.

tombruijn:
@jeffkreeftmeijer Thanks for your perspective. I had gotten the feeling it was approached that way. Personally I don't think this is mergeable. As discussed, can you take another pass at it to make it more human friendly and maintainable?

jeffkreeftmeijer:
I've updated the docs to remove most of the cruft, including some documentation files that only regurgitated the code, as well as the direct links to the code and most of the code snippets. I've also added a note about the decorators.

tombruijn:
@jeffkreeftmeijer you may need to push your changes.

docs/configuration-system.md
```diff
@@ -0,0 +1,160 @@
...
+- `send_params: true` - Request parameters are sent by default
+
+### 2. System Detection
+
+Automatically detected system configuration.
+If `APPSIGNAL_PUSH_API_KEY` is set, AppSignal is automatically activated.
+
+### 3. Application Configuration
+
+Configuration from `config/config.exs` or environment-specific config files:287-289):
```
tombruijn:
```suggestion
Configuration from `config/config.exs` or environment-specific config files:
```

docs/configuration-system.md
```diff
@@ -0,0 +1,160 @@
...
+All other configuration has sensible defaults, so the agent can start without them.
+Without a valid push API key, the agent cannot send data to AppSignal, making it pointless to run.
+
+## Active vs. Configured
+
+There's an important distinction between **configured as active** and **actually active**:
+
+### `configured_as_active?/0`
+
+Returns `true` if the `active` config option is set to `true` (lib/appsignal/config.ex:122-124):
```
tombruijn:
```suggestion
Returns `true` if the `active` config option is set to `true`:
```

docs/configuration-system.md
```diff
@@ -0,0 +1,160 @@
...
+### Default-Disabled Integrations
+
+- `enable_error_backend: false` - Error backend for crash reporting
+- `enable_minutely_probes: true` - System metrics (CPU, memory)
+- `enable_statsd: false` - StatsD server
+- `enable_nginx_metrics: false` - Nginx metrics collection
+
+## Special Configuration: Oban Error Reporting
+
+The `report_oban_errors` option has three possible values (lib/appsignal/config.ex:223-249):
```
tombruijn:
```suggestion
The `report_oban_errors` option has three possible values:
```

tombruijn:
I think this is better. Less noise. May improve upon this later.

PR #1035, opened 2026-03-21
Scope working_directory_path by otp_app

almirsarajcic:
When multiple Phoenix apps run on the same server, they all default to `/tmp/appsignal` and share one agent socket. The first app to boot "owns" the socket — all subsequent apps connect to it, sending their telemetry under the wrong API key.

This PR scopes `working_directory_path` to `<tmp>/appsignal-<otp_app>` when `otp_app` is set and no explicit path is configured. Apps in the cluster (multiple OS processes, same `otp_app`) still share one agent as intended.

Explicit `working_directory_path` or `working_dir_path` values always take precedence.

Related to #278 and #363.

### DIFF
--- a/lib/appsignal/config.ex
+++ b/lib/appsignal/config.ex
@@ -91,6 +91,18 @@ defmodule Appsignal.Config do
   defp determine_overrides(config) do
     %{}
     |> Map.merge(skip_session_data_backwards_compatibility(config, config[:skip_session_data]))
+    |> Map.merge(default_working_directory_path(config))
+  end
+
+  defp default_working_directory_path(config) do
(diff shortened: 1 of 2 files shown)


unflxw:
Hi @almirsarajcic, thank you for your proposed contribution!

Unfortunately, after considering your contribution, we have chosen to decline it.

It is not necessarily true that there's a one-to-one relation between the `otp_app` value and whether the same agent reports data for a given application. Manually configuring `otp_app` is not a very different experience from manually configuring `otp_app` _and_ `working_directory_path`.

Given the above, we'd rather continue to instruct customers to explicitly change `working_directory_path` when needed, rather than having some other configuration option implicitly change it, resulting in behaviour that works most of the time but leads to potentially confusing behaviour on edge cases.

PR #1039, opened 2026-05-12
Update Decimal dependency to 3.x

ccarvalho-eng:
Pins Decimal to `3.1.0`.

This is a dependency constraint update; I did not find direct Decimal API usage in the codebase.

Verification:
- `mix deps.get`
- `MIX_ENV=test_no_nif mix compile`

I also tried `MIX_ENV=test_no_nif mix test`, but the run is currently blocked by an existing Finch option compatibility error (`:ssl_options` / `:pool`) in the transmitter path before this Decimal change is exercised.

[skip changeset]

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -162,7 +162,7 @@ defmodule Appsignal.Mixfile do
     [
       {:castore, "~> 1.0"},
       {:certifi, "~> 2.14"},
-      {:decimal, "~> 3.0"},
+      {:decimal, "3.1.0"},
       {:benchee, "~> 1.0", only: :bench},
       {:finch, "~> 0.19"},
       {:jason, "~> 1.0"},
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -162,7 +162,7 @@ defmodule Appsignal.Mixfile do
     [
       {:castore, "~> 1.0"},
       {:certifi, "~> 2.14"},
-      {:decimal, "~> 2.0"},
+      {:decimal, "3.1.0"},
```
DerTim1: Should be `~> 3.1` or even better `~> 2.0 or ~> 3.1`.

ccarvalho-eng: Updated to `~> 2.0 or ~> 3.1`. I reran `mix deps.get` and `MIX_ENV=test_no_nif mix compile` successfully after the change.

lipskis (approved):

lipskis:
Thanks @ccarvalho-eng!

PR #1040, opened 2026-05-13
Fix Finch 0.22 strict option validation errors

nallwhy:
## Summary
- Move `:ssl_options` and `:proxy` from `Finch.request/3` options to pool-level `conn_opts` in `Finch.start_link`, fixing a compile-time crash with Finch 0.22+ in the agent installer and a runtime crash in the transmitter.
- Scope SSL `transport_opts` to HTTPS URLs so plain-HTTP requests (e.g. test endpoints behind Bypass) don't get TLS-only options passed to `:gen_tcp`.
- Parse the installer's proxy URL into Mint's `{scheme, host, port, opts}` tuple.

## Context
Finch 0.22.0 strictly validates the keyword list passed to `Finch.request/3` via `Keyword.validate!/2`, accepting only `[:pool_timeout, :receive_timeout, :request_timeout, :pool_strategy]`. Earlier Finch versions silently ignored unknown keys, so the `:ssl_options`/`:pool`/`:proxy` keys this integration has been passing at the request level have been no-ops ever since the Hackney → Finch migration. With 0.22 they raise `ArgumentError`:

```
** (ArgumentError) unknown keys [:ssl_options] in [ssl_options: [...]],
   the allowed keys are: [:pool_timeout, :receive_timeout, :request_timeout, :pool_strategy]
    mix_helpers.exs:247: Mix.Appsignal.Helper.do_download_file!/2
```

This breaks `mix deps.compile appsignal` for downstream users on Finch 0.22+, and would break runtime transmission as well.

### DIFF
--- a/lib/appsignal.ex
+++ b/lib/appsignal.ex
@@ -54,7 +54,7 @@ defmodule Appsignal do
       {Appsignal.Monitor, []},
       {Appsignal.Probes, []},
       {Appsignal.CheckIn.Scheduler, []},
-      {Finch, name: AppsignalFinch}
+      {Finch, name: AppsignalFinch, pools: %{default: Appsignal.Transmitter.pool_options()}}
     ]
 
     result = Supervisor.start_link(children, strategy: :one_for_one, name: Appsignal.Supervisor)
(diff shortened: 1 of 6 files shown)


lipskis:
Thanks @nallwhy for the contribution!

I made a few adjustments to this PR in #1042 and updated commits to align with our Git style guide.

lipskis:
Continuing this in #1042.

PR #1044, opened 2026-05-14
Bump `plug` version

lipskis:
Update `plug` version constraint to use `plug` `1.17` for Elixir < `1.12`.

[skip changeset]
[skip review]


lipskis:
This version bump doesn't fix the problem for Elixir 1.11. All of the new `plug` versions use `Kernel.then/2` which was added in Elixir 1.12.

PR #1055, opened 2026-09-09
Lock to mint to 1.9.2 on Elixir < 1.15

bjacquet:
Finch's latest version depends on Mint ~>1.8.  Mint versions until v1.9.1
require Elixir version ~>1.12.   Starting with Mint v1.9.2, it requires
Elixir ~>1.15.
The combination of this dependency specification forces us to set a
specific Mint version depending on the Elixir version.  Thus if the Elixir
version is lower than 1.15 we shall use Mint v1.9.2.

[skip changeset]

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -150,6 +150,15 @@ defmodule Appsignal.Mixfile do
         _ -> []
       end
 
+    mint_dependency =
+      case Version.compare(system_version, "1.15.0") do
+        :lt ->
+          [{:mint, "1.9.2"}]
+
+        _ ->
+          []
+      end
+
     plug_version =
       case Version.compare(system_version, "1.10.0") do
         :lt ->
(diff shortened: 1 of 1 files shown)


unflxw (approved):
LGTM! We should consider dropping very old Elixirs soon enough -- probably as part of the incoming collector mode work.

FYI, since this is your first Elixir release: make sure that you release from an Elixir/OTP version that is _not_ bound by these additional version requirements!

This is because the dependency requirements for the Elixir and OTP versions that is used to perform the release are "baked in" into the manifest for the released package, meaning we'd be forcing them on customers who don't necessarily need them. These additional requirement bounds are only for our CI.

unflxw:
@bjacquet After writing the above comment, I thought asking you to make sure you released from the right Elixir version was a bit silly, when we could assert that the right version was being used as part of the publishing process:

- https://github.com/appsignal/appsignal-elixir/pull/1056
- https://github.com/appsignal/appsignal-elixir-plug/pull/65
- https://github.com/appsignal/appsignal-elixir-phoenix/pull/126

PR #1058, opened 2026-09-18
Add Broadway instrumentation

bjacquet:
This adds `Appsignal.Broadway`, a minimal implementation that starts spans
for Broadway events.

Instruments the following Broadway telemetry events and created an
AppSignal child span:
- `[:broadway, :processor, :start]`
- `[:broadway, :processor, :stop]`
- `[:broadway, :processor, :message, :start]`
- `[:broadway, :processor, :message, :stop]`
- `[:broadway, :processor, :message, :exception]`


.changesets/add-broadway-instrumentation.md
```diff
@@ -0,0 +1,12 @@
+---
+bump: minor
+type: add
+integrations: elixir
```
unflxw, line 4:
```suggestion
```

This is only a thing in the agent repo, to route changesets to the right integrations on the auto-generated upgrade PR. It's harmless here, but it's not necessary.

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#prepare_messages")
+    |> @span.set_attribute("appsignal:category", "processor.broadway")
+    |> @span.set_attribute("index", index)
+    |> @span.set_attribute("message_count", length(messages))
+    |> @span.set_attribute("processor_key", to_string(processor_key))
+    |> @span.set_attribute("producer", module_name(producer_module))
+    |> @span.set_attribute("telemetry_span_context", inspect(telemetry_span_context))
```
unflxw, line 74:
```suggestion
```

The telemetry span context is not user-meaningful. It's a correlation value used to correlate the `:start` telemetry event with its corresponding `:stop` telemetry event.

Ideally we would use those to close the right stop event correlated with a specific start event, rather than simply closing the last one we opened. But I'd consider that change difficult due to how our tracer currently works, and therefore out of scope for this PR.

In any case, it's not meaningful to the customer and so there is no value in showing it to them.

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      messages: messages,
+      telemetry_span_context: telemetry_span_context,
+      producer: {producer_module, _}
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#prepare_messages")
+    |> @span.set_attribute("appsignal:category", "processor.broadway")
+    |> @span.set_attribute("index", index)
```
unflxw, line 70: Is the index user-meaningful? My guess is that it's not, but I lack the context to know for sure.

.changesets/add-broadway-instrumentation.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Broadway instrumentation is now available!  Module `Appsignal.Broadway` adds a minimal instruments covering the following Broadway telemetry events:
+- `[:broadway, :processor, :start]`
+- `[:broadway, :processor, :stop]`
+- `[:broadway, :processor, :message, :start]`
+- `[:broadway, :processor, :message, :stop]`
+- `[:broadway, :processor, :message, :exception]`
```
unflxw, line 12: The changelog should describe what the Broadway instrumentation does from a customer-facing perspective. The Broadway telemetry events and the module that handles them are an implementation detail, as the customer does not at any point interact with the module.

```suggestion
Automatically instrument Broadway applications.

When your Broadway application's processor processes one or more messages, a trace will be created to represent the processing of these messages. The root span encompasses the call to the `prepare_messages/2` callback, if defined, and each of its child spans will represent the individual processing of each of these messages through the `handle_message/3` callback.
```

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      name: name,
+      processor_key: processor_key,
+      index: index,
+      messages: messages,
+      telemetry_span_context: telemetry_span_context,
+      producer: {producer_module, _}
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
```
unflxw, line 67: I believe we'd always want the processor span to be a root span (the start of a new trace) and never a child span.

```suggestion
    do_broadway_processor_start(metadata)
  end

  defp do_broadway_processor_start(metadata) do
    %{
      topology_name: topology_name,
      name: name,
      processor_key: processor_key,
      index: index,
      messages: messages,
      telemetry_span_context: telemetry_span_context,
      producer: {producer_module, _}
    } = metadata

    "broadway"
    |> @tracer.create_span()
```

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+    } = metadata
+
+    @tracer.current_span()
+    |> @span.set_attribute("failed_messages", length(failed_messages))
+    |> @span.set_attribute("successful_messages_to_ack", length(to_ack))
+    |> @span.set_attribute("successful_messages_to_forward", length(to_forward))
+    |> @span.set_attribute(
+      "duration_ms",
+      System.convert_time_unit(duration, :native, :millisecond)
+    )
```
unflxw, line 92:
```suggestion
```

This is not an useful attribute, as it is redundant with the duration of the span itself. (Let me know if I'm missing context by which it's not!)

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#prepare_messages")
+    |> @span.set_attribute("appsignal:category", "processor.broadway")
+    |> @span.set_attribute("index", index)
+    |> @span.set_attribute("message_count", length(messages))
+    |> @span.set_attribute("processor_key", to_string(processor_key))
+    |> @span.set_attribute("producer", module_name(producer_module))
```
unflxw, line 73: What's the difference between `name` (which we use as `module_name(name)` in the span name) and `producer_module` here? If they're the same, let's not report both. If `name` is an arbitrary user-defined name (for example given in the topology config) and `producer_module` is the actual implementer module, then I believe it is `producer_module` that should be used in the span name.

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      producer: {producer_module, _}
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#prepare_messages")
+    |> @span.set_attribute("appsignal:category", "processor.broadway")
+    |> @span.set_attribute("index", index)
+    |> @span.set_attribute("message_count", length(messages))
+    |> @span.set_attribute("processor_key", to_string(processor_key))
```
unflxw, line 72: Same question re: user meaningfulness for `processor_key`. Again I may be lacking context here of how these map from Broadway concepts.

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      telemetry_span_context: telemetry_span_context,
+      producer: {producer_module, _}
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#prepare_messages")
+    |> @span.set_attribute("appsignal:category", "processor.broadway")
+    |> @span.set_attribute("index", index)
+    |> @span.set_attribute("message_count", length(messages))
```
unflxw, line 71: `message_count` could be an useful metric, if we choose to implement those!

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      producer: {producer_module, _}
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(@tracer.current_span())
+    |> @span.set_name("#{module_name(name)}#handle_message")
+    |> @span.set_attribute("appsignal:category", "processor_message.broadway")
+    |> @span.set_attribute("index", index)
+    |> @span.set_attribute("processor_key", to_string(processor_key))
+    |> @span.set_attribute("producer", module_name(producer_module))
```
unflxw, line 114: My understanding (pending testing this with the Broadway test setup) is that these `processor_message.broadway` spans would be child spans of the main `processor.broadway` span.

This may have changed after the legacy-to-OpenTelemetry translation layer, in which case, it is a bug in that layer, although we may wish to promote this bug to feature. But in the legacy AppSignal spans API behaviour (in a broad sense of the term, including the legacy data processing pipeline and the legacy UI for samples that it fed into) attributes in child spans (other than `appsignal:category` and `appsignal:body`) would not be stored or shown in any way. Attributes were only allowed in root spans, where they were essentially promoted into tags (as in what the `set_tags` helper does in most other integrations)

If these attributes are now being shown, my gut feeling is that we should fix these bugs in the legacy-to-OpenTelemetry translation layer for the spans API, as to not unintentionally change the established contract. But this is not a strongly held opinion and I could be convinced otherwise.

I'm also unsure which of these attributes would are user-meaningful. `processor_key` would be redundant with the root span's own processor, no?

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+  defp do_broadway_processor_start(metadata, current_span) do
+    %{
+      topology_name: topology_name,
+      name: name,
+      processor_key: processor_key,
+      index: index,
+      messages: messages,
+      telemetry_span_context: telemetry_span_context,
+      producer: {producer_module, _}
+    } = metadata
```
unflxw, line 64: This pattern matching would not match [in older versions of Broadway, where `producer` is not present](https://broadway.hexdocs.pm/1.0.0/Broadway.html#module-telemetry), causing the telemetry handler to crash.

As Telemetry does not give us the granularity to know which version of Broadway is emitting these events, we want to only use pattern matching (if at all) if and only if we know that a given metadata key was always emitted, and gate access to the optionally-present metadata using something like `if metadata[:producer]` otherwise.

For specific examples, see how our [Finch](https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/finch.ex#L31-L56) and [Oban](https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/oban.ex#L81-L106) instrumentations handle it: Finch is an example of a scenario where the shape is too different, so we detect it through an absence of match and intentionally do not support it (this is in turn explicitly stated in our docs) while Oban is an example of a scenario where we pattern-match on the common denominator, then augment its behaviour for versions that carry more data. This specific instance of it looks more like an Oban-shaped scenario than a Finch-shaped one to me, though it is ultimately your call.

(I'm pointing out a specific instance of this issue in a specific Broadway version and a specific handler, but we should check all handlers in all versions that emitted telemetry events with the same naming patterns for potential variations of this issue)

lib/appsignal/broadway.ex
```diff
@@ -0,0 +1,145 @@
...
+      name: name,
+      reason: reason,
+      stacktrace: stacktrace
+    } = metadata
+
+    "broadway"
+    |> @tracer.create_span(current_span)
+    |> @span.set_name("#{module_name(name)}#handle_message")
+    |> @span.set_attribute("appsignal:category", "exception.broadway")
+    |> @span.add_error(kind, reason, stacktrace)
```
unflxw, line 142: Linking to my other message for context as to not repeat myself: https://github.com/appsignal/appsignal-elixir/pull/1058/changes#r4075482918

Same as with attributes in the other message, but for errors: errors in child spans were not supported before the legacy-to-OpenTelemetry API, where they were ignored entirely; if they are now, it's almost certainly a bug in terms of the correctness of the translation, although it is a bug that we may choose to promote to feature.

Beyond whether they were supported, the spans API also did not support more than one error in the same _sample_ (which is now a _trace_, thanks to the legacy-to-OpenTelemetry translation layer) and so we should check what happens if several spans error in the same trace -- is only one of them reported, or are all of them reported?

Like with the other comment, I'd rather "fix the bug" and rectify the legacy-to-OpenTelemetry translation semantics to what they originally were in the legacy systems, and then we can choose to do cooler things with collector mode and OpenTelemetry support later on, rather than embrace these "accidental improvements" (if they do indeed work correctly -- like with the other message, I haven't checked just yet)

unflxw (changes requested):
Good starting point! I think it needs some polish in the details, as described in the review, but the core is solid.

Something I wasn't unsure of where to mention: in the current implementation, a whole batch of messages to process is the root span (`processor.broadway`) and the processing that takes place for each of those individual messages (`processor_message.broadway` spans, which encompass calls to the customer's `handle_message/3` callback) are child spans within it.

I'm not sure if this is a good idea. This is not a super strongly held opinion, but I do lean the other way, toward `processor.broadway` not being instrumented (or somehow only instrumenting `process_messages/2` separately?) and instrumenting `processor_message.broadway` as the root span, allowing the customer's work (custom instrumentation, calls to Ecto, etc) to take place within it.

The reason being my belief (perhaps uninformed in the ways of Broadway) that the meaningful unit of work here is the call to `handle_message/3`, where each message is processed individually, and where this processing can fail individually, and not so much the batch call through `process_messages/2`. If each message could result in an error being reported to AppSignal, it's a nicer experience if each error is reported separately and leads to a trace about the handling of the specific message that triggered the error, rather than error events in a broader trace that did not fail. In terms of the resulting metrics for throughput and so on it's also nicer if the throughput and the error rate matches the number of messages that were reported, rather than the number of batches that were reported.

lib/appsignal/broadway.ex
bjacquet: Broadway's documentation doesn't clarify what this _index_ is, apart from being a positive integer, but other parts of the docs refer to _index_ as a producer's positional index in Broadway's supervision tree.  
I don't think it's very useful for users when they're looking for the cause of issues; I'd say it is mostly informative.

lib/appsignal/broadway.ex
bjacquet: I was thinking we might want to "link" this new span to another (open) span that already exists (from other instrumentation).  Maybe it made sense to [do that for Absinthe](https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/absinthe.ex#L48) but not for Broadway?  We also [don't do it for Oban](https://github.com/appsignal/appsignal-elixir/blob/main/lib/appsignal/oban.ex#L70).

lib/appsignal/broadway.ex
unflxw: Ah, also, we shouldn't create a new span for the exception, but add the error to the existing span.

lib/appsignal/broadway.ex
unflxw: I'm not sure why we'd do that for Absinthe, it's not an instrumentation I'm super familiar with. I think Absinthe can be called from within Plug and Phoenix, going by its docs, so it would make sense to nest it within the Plug and Phoenix context.

The rule of thumb is whether this represents the start of a new self-contained of work (handling an HTTP response, processing a background job) or if it's something that takes place during other work (performing an HTTP request, fetching data from the database, enqueueing a background job). I believe Broadway's processor work would fall on the first category.

lib/appsignal/broadway.ex
unflxw: Looked into it using the test setup: `name` is a name representing the specific instance of the processor, part of which is auto-generated (so, with concurrency set to `4`, you get `BroadwayExample.Pipeline.Broadway.Processor_default_0` to `BroadwayExample.Pipeline.Broadway.Processor_default_0`) while it's `topology_name` (`BroadwayExample.Pipeline`) that we'd want to use instead. 

The `producer` is its own meaningful value describing where the message came from -- my bad, I mistook "processor" and "producer" 🤦

`processor_key` is seemingly always `default`, so I think that's likely just noise that we should also remove.

bjacquet: `name` is the processor's (instance) name, e.g. `BroadwayExample.Pipeline.Broadway.Processor_default_0`, and `producer_module` is the actual implementer module, e.g. `BroadwayExample.Producer`.  
These module's name examples are taken from https://github.com/appsignal/test-setups/pull/387.

unflxw: Note that both `name` and `topology_name` can be `{:via, module, term}` tuples instead of plain modules. We should deal with that in some way (even if that way is just `inspect()` -- there may be code that deals with it more nicely elsewhere, though maybe not) rather than crashing.

lib/appsignal/broadway.ex
unflxw: Informative of what? If we don't know what it's informative of, let's drop it rather than adding more noise for the customer. (It's the concurrency index, see the other comment re: concurrency -- not useful information, in my opinion)

lib/appsignal/broadway.ex
unflxw: Checked against the pre-translation behaviour, I was wrong: errors in child spans _were_ supported (by copying them to the root span; they still are post-translation) but we only report one of them at a time, both pre- and post-translation.

To me this argues in the direction of my review-wide comment (instrument each `handle_message/3`/`:processor_message` event as its own root span rather than nested within `process_messages/2`/`:processor`) in that, as currently implemented, if more than one exception occurs, we'll only report one of them.

lib/appsignal/broadway.ex
unflxw: > name is the processor's (instance) name, e.g. BroadwayExample.Pipeline.Broadway.Processor_default_0, and producer_module is the actual implementer module, e.g. BroadwayExample.Producer.

Yes, indeed. The root span name will become the action name and be used to group similar traces together. The `_0` at the end is the concurrency index.

We don't want to group events processed by concurrency index 2 separately from events by concurrency index 1, as they're all, conceptually, events processed by the same processor. 

Therefore we should (I think? there may be better candidates I'm not aware of) use `topology_name`, which would just be `BroadwayExample.Pipeline`.

lib/appsignal/broadway.ex
unflxw: Tested this with the test setup, these attributes are indeed dropped both pre- and post- translation layer, so there's no point in setting them.

```suggestion
    |> @span.set_attribute("appsignal:category", "processor_message.broadway")
```

lib/appsignal/broadway.ex
bjacquet: `processor_key` is a key (atom) that identifies the processor group, where users define their Broadway topology.  Although users define a list of processor groups, it [only allows one group](https://github.com/elixir-broadway/broadway/blob/main/lib/broadway/topology.ex#L267-L271) to be defined.  
Until Broadway supports multiple groups, this is mostly informative.

lib/appsignal/broadway.ex
bjacquet: Don't know what _more_ you want to implement.  `messages` is a list and may contain more than one message.  If I understood Broadway correctly, a processor receives this list of messages and dispatches them to producers.

unflxw: No, it's good. Right now it's an attribute on a trace. What I meant to say was that, in the future, if we wanted to implement Broadway _metrics_ (alongside the currently implemented _traces_) then `message_count` (and also the `to_ack`, `to_forward` values elsewhere) could be good candidates for metrics.

It was just a happy thought, completely out-of-scope for the PR at this moment, not homework.

lib/appsignal/broadway.ex
bjacquet: In my best intentions and naivety I was assuming these attributes would show up in the span details bar.  As you say, the attributes only show up for the root span.  

If indeed there's no way to show attributes for child spans then I suggest to remove these lines, which do nothing.

unflxw: Yep, unfortunately no way to do that just yet, because of how our legacy spans API works. We can bring them back once we have collector mode, which will enable some way of setting attributes on child spans.

.changesets/add-broadway-instrumentation.md
bjacquet: Removed.

lib/appsignal/broadway.ex
bjacquet: Removed.

lib/appsignal/broadway.ex
bjacquet: Removed.

.changesets/add-broadway-instrumentation.md
bjacquet: Updated.

lib/appsignal/broadway.ex
bjacquet: Yes, Broadway doesn't react to external stimuli to commence work.

Applied suggestion.

lib/appsignal/broadway.ex
bjacquet: I changed things a bit so that it will be:
- `name`: `BroadwayExample.Producer#prepare_message`
- `producer`: `BroadwayExample.Producer`
- `processor`: `BroadwayExample.Pipeline.Broadway.Processor_default_1`
- `topology_name`: `BroadwayExample.Pipeline`

Let me know what you think.

lib/appsignal/broadway.ex
bjacquet: Removed `processor_key` and added `processor`.  See [this comment](https://github.com/appsignal/appsignal-elixir/pull/1058#discussion_r4085190997).

lib/appsignal/broadway.ex
bjacquet: Gotcha!  👍🏼  Keeping `message_count` then.

lib/appsignal/broadway.ex
bjacquet: Removed everything except `name` and `appsignal:category`.

lib/appsignal/broadway.ex
bjacquet: Changed implementation to use the `metadata[:attribute]` pattern instead.

lib/appsignal/broadway.ex
bjacquet: Ohh!  Now I don't know if we oughta create a new error span or not.  😅 

In the meantime, I removed the new span creation.

lib/appsignal/broadway.ex
bjacquet: Removed this attribute.

